PR #53, opened 2023-02-08
fix: replace uglifier by terser for js compression

nicolas-brousse:
This PR replace `uglifier` gem by `terser` for js compression.


nicolas-brousse:
@nanego if it's good for you, you can merge this.

PR #96, opened 2023-10-13
275817 - i18n - Disk types resource

MBausson:


nicolas-brousse (changes requested):
@MBausson some change to do

nicolas-brousse (approved):

PR #97, opened 2023-10-18
Update ApplicationController to allow API requests

B-Rass:


app/controllers/application_controller.rb
```diff
@@ -10,7 +10,7 @@ class ApplicationController < ActionController::Base
 
   # Prevent CSRF attacks by raising an exception.
   # For APIs, you may want to use :null_session instead.
-  protect_from_forgery with: :exception
+  protect_from_forgery with: :exception, unless: -> { request.format.json? }
```
nicolas-brousse, line 13:
```suggestion
  protect_from_forgery with: :exception, unless: -> { request.format.json? }
  protect_from_forgery with: :null_session, if: -> { request.format.json? }
```

nicolas-brousse (changes requested):

nicolas-brousse (approved):

PR #132, opened 2024-01-24
294488 - API : Add new use-cases

B-Rass:


app/controllers/servers_controller.rb
```diff
@@ -7,6 +7,7 @@ class ServersController < ApplicationController
 
   def index
     @servers = Server.includes(:frame, :room, :islet, bay: :frames, modele: :category).order(:name)
+    @servers = @servers.where(name: params[:name]) if params[:name].present?
```
nicolas-brousse, line 10:
```suggestion
    @servers = @servers.where(Server.arel_table[:name].matches("%#{params[:name]}%")) if params[:name].present?
```

I suggest the usage of `LIKE` SQL, to let user search with partial name.

B-Rass: Indeed, thanks

nicolas-brousse (approved):

PR #151, opened 2024-02-28
Fix: don't use user#authentication_token.length for hidden value

nicolas-brousse:


app/views/users/show.html.erb
```diff
@@ -35,7 +35,7 @@
               <%= @user.authentication_token %>
             </span>
             <span data-reveal-target="item">
-              <%= "*" * @user.authentication_token.length %>
```
B-Rass, line 38: Why setting a default value (20) and not `0`?

nicolas-brousse: It looked weird. Maybe the better should be to simple show empty value and hidden reveal too

B-Rass: yes

nicolas-brousse: Updated

B-Rass (approved):

PR #152, opened 2024-02-28
Rubocop cleanups

nicolas-brousse:


app/views/users/show.html.erb
B-Rass, line 1: Should not be here

nicolas-brousse: Fixed

B-Rass (approved):

PR #156, opened 2024-03-06
Update changelog index

nicolas-brousse:


spec/decorators/changelog_entry_decorator_spec.rb
B-Rass, line 1: In a perfect world, it may be nice to add the methods test

PR #157, opened 2024-03-20
Use datatable in users index

nicolas-brousse:


app/components/list/data_table_component.rb
```diff
@@ -59,8 +59,8 @@ def link_to_sort(label, attribute)
       current_direction = current_attribute == attribute ? params[:direction].to_sym : nil
 
       parameters = case current_direction
-                   when nil then { column: attribute, direction: :asc }
-                   when :asc then { column: attribute, direction: :desc }
+                   when nil then { column: attribute, direction: :desc }
+                   when :desc then { column: attribute, direction: :asc }
```
B-Rass, line 63:
```suggestion
                   when nil then { column: attribute, direction: :asc }
                   when :asc then { column: attribute, direction: :desc }
```

B-Rass: In fact, keep it like it was, it should order a to z first.

B-Rass (changes requested):
My bad about the asc / desc changes

app/components/list/data_table_component.rb
nicolas-brousse: @B-Rass I changed it back

B-Rass (approved):

PR #171, opened 2024-04-08
Create card component

MBausson:


spec/components/previews/card_component_preview.rb
nicolas-brousse, line 1: May be nice to have header text and footer text customizable in lookbook

app/components/card_component.rb
nicolas-brousse, line 1: Could be nice to already be able to set a color (see.: https://getbootstrap.com/docs/3.4/components/#panels-alternatives)

nicolas-brousse (changes requested):
@MBausson good start, I see two part that could be improve

app/components/card_component.rb
```diff
@@ -0,0 +1,33 @@
...
+      <% if footer? %>
+      <div class="panel-footer">
+        <%= footer %>
+      </div>
+      <% end %>
+    </div>
+  ERB
+
+  def initialize(state = :default)
+    @state = state
```
nicolas-brousse, line 29:
```suggestion
    @type = type
```

Perfer type over state.

Also add list of available types in a constant. And if value is not in the list then raise an `ArgumentError` exception:

```rb
raise ArgumentError, "#{type} is not a valid type" unless TYPES.key?(type)
```

spec/components/card_component_spec.rb
nicolas-brousse, line 1: Add test for the `type` argument

- 1 that works and check presence of class
- 1 that doesn't works and throw an exception

spec/components/previews/card_component_preview.rb
nicolas-brousse: Rename `state` into `type` here too.

nicolas-brousse (changes requested):

nicolas-brousse (approved):

PR #176, opened 2024-04-17
Fixing wrong network data migration

B-Rass:
Migration network data was done badly. Here's the fix of it


B-Rass:
I compared capacity and infrastructure views before / after executing this migration, and I had the same results.

I also check on both DBs and I counted same amount for each network type. So, the fix is valid to me.

nicolas-brousse (approved):

PR #181, opened 2024-05-02
320594 Store user settings

MBausson:


db/migrate/20240502093805_enable_hstore.rb
nicolas-brousse, line 1: Is this pg extension necessary?

app/views/users/settings/edit.html.erb
```diff
@@ -0,0 +1,30 @@
+<div class="authform">
+  <h3><%= t(".title") %></h3>
+
+  <%= form_for @user, url: users_settings_path do |f| %>
```
nicolas-brousse, line 4:
```suggestion
  <%= form_with model: @user, url: users_settings_path do |f| %>
```

app/views/users/settings/edit.html.erb
```diff
@@ -0,0 +1,30 @@
...
+    <div class="form-group">
+      <%= f.label :locale %>
+      <%= f.select :locale, User::AVAILABLE_LOCALES.compact, selected: @user.locale, class: "form-control" %>
```
nicolas-brousse, line 7: `selected` should not be necessary if form configured properly

app/views/users/settings/edit.html.erb
```diff
@@ -0,0 +1,30 @@
...
+    </div>
+
+    <!-- TODO: Remove me when themes will be implemented -->
+<!--    <div class="form-group">-->
+      <%#= f.label :theme %>
+      <%#= f.select :theme, User::AVAILABLE_THEMES.compact, selected: @user.theme, class: "form-control" %>
```
nicolas-brousse, line 13: `selected` should not be necessary if form configured properly

app/views/users/settings/edit.html.erb
```diff
@@ -0,0 +1,30 @@
...
+<!--    </div>-->
+
+    <div class="form-group">
+      <%= f.label :visualization_bay_default_background_color %>
+      <%= f.select :visualization_bay_default_background_color, User::AVAILABLE_BAY_BACKGROUND_COLORS.compact,
+                   selected: @user.visualization_bay_default_background_color, class: "form-control" %>
```
nicolas-brousse, line 19: `selected` should not be necessary if form configured properly

app/views/users/settings/edit.html.erb
```diff
@@ -0,0 +1,30 @@
...
+    </div>
+
+    <div class="form-group">
+      <%= f.label :visualization_bay_default_orientation %>
+      <%= f.select :visualization_bay_default_orientation, User::AVAILABLE_BAY_ORIENTATIONS.compact,
+                   selected: @user.visualization_bay_default_orientation, class: "form-control" %>
```
nicolas-brousse, line 25: `selected` should not be necessary if form configured properly

app/views/users/settings/edit.html.erb
nicolas-brousse, line 7: What's `compact` used for?

app/models/user.rb
```diff
@@ -1,9 +1,21 @@
 # frozen_string_literal: true
 
 class User < ApplicationRecord
+  AVAILABLE_LOCALES = (I18n.available_locales.map(&:to_s) << nil).freeze
```
nicolas-brousse, line 4:
```suggestion
  AVAILABLE_LOCALES = I18n.available_locales.map(&:to_s).freeze
```

No needs of `nil`. You have to use `allow_nil: true` in validator (apply the same on every validators related to settings

db/migrate/20240502093805_enable_hstore.rb
MBausson: No, i removed this migration

app/views/users/settings/edit.html.erb
MBausson: Since every constants had a nil value, I wanted not to show it on views
The `compact` is not necessary anymore

app/views/users/settings/edit.html.erb
```diff
@@ -0,0 +1,28 @@
+<div class="authform">
+  <h3><%= t(".title") %></h3>
+
+  <%= form_with model: @user, url: users_settings_path do |f| %>
+    <div class="form-group">
+      <%= f.label :locale %>
+      <%= f.select :locale, User::AVAILABLE_LOCALES, class: "form-control" %>
```
MBausson, line 7: @nicolas-brousse Do you have an idea of why the class is not being applied on the select ?

B-Rass (approved):

B-Rass:
@nicolas-brousse I don't know if we merge this, because if an user choose `english` as his locale, then the site will crash, and will not be able to update it. 


+ I let you review my small changes

app/controllers/concerns/localization.rb
```diff
@@ -0,0 +1,15 @@
...
+module Localization
+  extend ActiveSupport::Concern
+
+  included do
+    before_action :set_locale
+
+    private
+
+    def set_locale
+      I18n.locale = current_user&.locale || I18n.default_locale.to_s
```
nicolas-brousse, line 12: Does `to_s` necessary?

config/locales/en.yml
nicolas-brousse, line 1: Add missing translations

config/locales/activerecord.fr.yml
nicolas-brousse, line 1: Add this tranlations in `en` file

app/views/rooms/show.html.erb
```diff
@@ -71,7 +73,7 @@
                                 <% u_disponibles -= 1 %>
                               <% end %>
 
-                              <% bgModeleColor = define_background_color(server: server, mode: params['bg']) %>
+                              <% bgModeleColor = define_background_color(server: server, mode: bg || Modele.model_name.human.downcase) %>
```
nicolas-brousse, line 76: For which usage `Modele.model_name.human.downcase` has been added?

app/views/rooms/show.html.erb
```diff
@@ -1,3 +1,5 @@
+<% bg = params[:bg].presence || current_user.visualization_bay_default_background_color %>
+<% view = params[:view].presence || current_user.visualization_bay_default_orientation == 'rear' ? 'back' : 'front' %>
```
nicolas-brousse, line 2: Is there any reason for using `rear` in our settings instead `back`?

app/views/rooms/show.html.erb
```diff
@@ -114,7 +116,7 @@
         <!-- Choix de la couleur de fond des serveurs -->
         <div class="btn-group">
           <button type="button" class="btn btn-default dropdown-toggle" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false">
-            Couleur de fond : <%= params[:bg] || 'modèle' %> <span class="caret"></span>
+            <%= t("rooms.action_buttons.background_color", bg: bg || Modele.model_name.human.downcase) %> <span class="caret"></span>
```
nicolas-brousse, line 119: Not sure about using `Modele.model_name.human.downcase`. It seams to be a key here, so if we rename Modele model name translation, this will break this code.

app/views/rooms/show.html.erb
```diff
@@ -206,20 +208,19 @@
...
-        if (searchParams.has('view')) {
-            view = searchParams.get('view')
-        } else {
-            searchParams.set('view', view)
+        const searchParams = new URLSearchParams(window.location.search)
+
+        if (!searchParams.has('view')){
+          searchParams.set('view', '<%= view %>')
         }
+
```
nicolas-brousse, line 216: Is this still necessary? `searchParams` seams not to be used anymore

app/views/users/settings/edit.html.erb
```diff
@@ -0,0 +1,41 @@
+<div class="authform">
+  <h3><%= t(".title") %></h3>
+
+  <%= form_with model: @user, url: users_settings_path do |f| %>
+    <div class="form-group">
+      <%= f.label :locale %>
+      <%= f.select :locale,
+                   User::AVAILABLE_LOCALES.map { |locale| [User.human_attribute_name("locale.#{locale}"), locale] },
```
nicolas-brousse, line 8: Move this in a decorator.

For example:
```rb
class << self
    def available_locales_for_options
      AVAILABLE_LOCALES.map { |locale| [User.human_attribute_name("locale.#{locale}"), locale] }
    end
end
```

nicolas-brousse: Also do the same for the others select

nicolas-brousse (changes requested):
@MBausson here is my review.

app/views/rooms/show.html.erb
MBausson: No, I actually used the same values as in the issue

app/views/users/settings/edit.html.erb
```diff
@@ -0,0 +1,28 @@
...
+  <%= form_with model: @user, url: users_settings_path do |f| %>
+    <div class="form-group">
+      <%= f.label :locale %>
+      <%= f.select :locale, UserDecorator::available_background_colors_for_options, {}, class: "form-control" %>
+    </div>
+
+    <!-- TODO: Remove me when themes will be implemented -->
+<!--    <div class="form-group">-->
+      <%#= f.label :theme %>
+      <%#= f.select :theme, User::AVAILABLE_THEMES.compact, class: "form-control" %>
```
nicolas-brousse, line 13: Move this in the decorator too.

app/decorators/user_decorator.rb
nicolas-brousse, line 1: Add decorator tests

nicolas-brousse (changes requested):
@MBausson 2 last changes otherwise LGTM

app/views/users/settings/edit.html.erb
```diff
@@ -0,0 +1,28 @@
+<div class="authform">
+  <h3><%= t(".title") %></h3>
+
+  <%= form_with model: @user, url: users_settings_path do |f| %>
+    <div class="form-group">
+      <%= f.label :locale %>
+      <%= f.select :locale, UserDecorator::available_background_colors_for_options, {}, class: "form-control" %>
```
nicolas-brousse, line 7: Not the good options.

nicolas-brousse (changes requested):

nicolas-brousse (approved):

PR #196, opened 2024-07-01
Add feature test for RoomsController#show

MBausson:


test/controllers/frames_controller_test.rb
```diff
@@ -21,7 +21,7 @@ class FramesControllerTest < ActionController::TestCase
 
   test "should create frame" do
     assert_difference('Frame.count') do
-      post :create, params: { frame: { name: @frame.name, bay_id: 1 } }
+      post :create, params: { frame: { name: "MyFrame4", bay_id: 1 } }
```
MBausson, line 24: Since I added slugs to frames fixtures, the test failed because of a unique constraint violation on slugs

Gemfile
```diff
@@ -70,7 +71,7 @@ group :test do
   gem "simplecov", require: false
   gem "simplecov-console", require: false
   # gem "minitest-rails-capybara"
-  # gem "formulaic"
+  gem "formulaic"
```
B-Rass, line 74: Can you put this line where it belongs, after `cuprite` line

B-Rass (changes requested):

spec/features/rooms/browse_room_spec.rb
```diff
@@ -0,0 +1,73 @@
...
+    end
+
+    # 3. Change background color to 'gestionnaire'
+    bg_dropdown_button = first("span.pull-right > * > .btn-default.dropdown-toggle")
+    bg_dropdown_button.click
+
+    bg_dropdown_gestionnaire = all(".dropdown-menu > li")[1]
+    bg_dropdown_gestionnaire.click
+
+    sleep 1
```
nicolas-brousse, line 64: What this for?

MBausson: Since the view is loaded via javascript, we cannot use `find()` to wait for the DOM to be updated because this is already done earlier in the test

B-Rass (approved):

nicolas-brousse (approved):

PR #202, opened 2024-09-05
340750 add islets in visualization dropdown

B-Rass:


app/views/layouts/navigation/_visualization_dropdown.html.erb
```diff
@@ -16,6 +22,17 @@
         <li class="nav-item" role="menuitem">
           <%= link_to room.name, room_path(room, view: params[:view], bg: params[:bg]), class: "dropdown-item" %>
         </li>
+        <% islets = room.islets.sorted.not_empty.has_name.distinct %>
+        <% if islets.size > 1 %>
+          <% islets.each do |islet| %>
```
nicolas-brousse, line 27:
```suggestion
          <% room.islets.sorted.not_empty.has_name.distinct.each do |islet| %>
```

Not sure we need to do a condition since there is no extra html tag to show if value or hide if no values

B-Rass: To test, but all room have islets, and I don't want to show it if only 1. But maybe `has_name` and `not_empty` doing the job here. 

To test to be sure

nicolas-brousse (approved):

PR #208, opened 2024-09-10
341607 Improve show pages

B-Rass:


app/components/page/heading_component.rb
B-Rass, line 1: @nicolas-brousse I need your thoughts on this

app/views/sites/index.html.erb
```diff
@@ -1,48 +1,48 @@
...
-<div class="container-fluid">
-  <div class="d-flex justify-content-between">
-    <h1><%= t(".title") %></h1>
-    <%= link_to new_site_path, class: "btn btn-success align-self-center" do %>
-      <span class="bi bi-plus-lg"></span>
-      <%= t(".add_site") %>
-    <% end %>
-  </div>
+<div class="p-4 bg-body">
+  <div id="map" class="border rounded mb-3" style="height: 450px;"></div>
```
nicolas-brousse, line 15: I think I prefer map outside from header block

app/views/sites/_form.html.erb
```diff
@@ -11,34 +15,46 @@
...
+          <%= f.label :longitude, class: "form-label" %>
+          <%= f.text_field :longitude, class: "form-control" %>
+        </fieldset>
+      </div>
+    <% end %>
+
+    <div class="col-12 py-3 text-end sticky-bottom bg-body-tertiary">
+      <%= link_to t("action.cancel"), sites_path, class: "btn btn-outline-secondary" %>
+      <%= f.submit class: "btn btn-primary" %>
     </div>
```
nicolas-brousse, line 58: I think a light top border may be better, and maybe a different background color

app/views/servers/index.html.erb
```diff
@@ -3,174 +3,172 @@
...
+        <%= f.collection_select(:cluster_id, Cluster.sorted,
+                                          :id,
+                                          :name,
+                                          { prompt: true, include_blank: true },
+                                          { class: "form-select" }) %>
+        <%= f.label :cluster_id %>
+      </fieldset>
+    </div>
   <% end %>
+<% end %>
```
nicolas-brousse, line 121: Not sure about having the filter in the header block

app/views/servers/edit.html.erb
nicolas-brousse, line 1: Is it deliberate to not having the form inputs in blocks?

app/views/sites/_form.html.erb
B-Rass: I'm not sure about another background color, because it's not full width. I was looking to add a border, and have it dynamic with the scroll

app/views/servers/edit.html.erb
B-Rass: Yes I don't know if I have to do it here also. You think I should ? Make it looks like on other page ?

I don't have name for all the blocks

spec/processors/servers_processor_spec.rb
```diff
@@ -295,6 +295,21 @@
...
+
+    before do
+      server
+      Server.create!(name: "server2", numero: 2, **attributes)
+    end
+
+    it { expect(result.size).to eq(1) }
+    it { is_expected.to include(server) }
+  end
+
```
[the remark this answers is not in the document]
B-Rass: done

app/views/air_conditioners/index.html.erb
nicolas-brousse, line 1: Move name column in first position

nicolas-brousse (approved):

PR #210, opened 2024-09-11
341617 - Add ChangelogEntry to show pages

MBausson:
Requires #205 to be merged first


app/helpers/changelog_entries_helper.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+module ChangelogEntriesHelper
+  def object_changelog_entries(object, **)
+    tag.turbo_frame(id: "changelog-entries", loading: :lazy, src: object_changelog_entries_path(object), **)
```
MBausson, line 5: Would it be interesting to add a sort of "loading" content to show while loading the turbo frame ?

nicolas-brousse: Yes. Also it may be better to use decorator instead helper

app/helpers/changelog_entries_helper.rb
```diff
@@ -0,0 +1,13 @@
...
+  end
+
+  private
+
+  def object_changelog_entries_path(object)
+    "/#{object.class.name.underscore.pluralize}/#{object.id}/changelog_entries"
```
nicolas-brousse, line 11: We should found a better way

app/helpers/changelog_entries_helper.rb
nicolas-brousse, line 5: Use turbo frame helper https://github.com/hotwired/turbo-rails/blob/main/app/helpers/turbo/frames_helper.rb

app/views/changelog_entries/index.html.erb
```diff
@@ -3,50 +3,52 @@
   t("changelog_entries.title") => ""
 } %>
 
-<div class="container-fluid">
-  <h1><%= t("changelog_entries.title") %></h1>
+<turbo-frame id="changelog-entries">
```
nicolas-brousse, line 6: Use turbo frame helper https://github.com/hotwired/turbo-rails/blob/main/app/helpers/turbo/frames_helper.rb

nicolas-brousse (changes requested):

nicolas-brousse (changes requested):

nicolas-brousse:
@B-Rass if it's good for you, then it may be apply on every pages.

nicolas-brousse (approved):

B-Rass (approved):

PR #216, opened 2024-09-18
Make Visualization controllers hide sidebar by default

nicolas-brousse:


app/controllers/visualization/base_controller.rb
```diff
@@ -0,0 +1,7 @@
+module Visualization
```
B-Rass, line 1:
```suggestion
# frozen_string_literal: true

module Visualization
```

B-Rass (changes requested):
One small change but otherwhile good to be merged

PR #218, opened 2024-09-18
308778 - Rename Composant model

nicolas-brousse:
- [ ] Found a better name than `EnclosureComponent` try not to use `Component` in name to avoid collisions with view components


B-Rass:
> * [ ]  Found a better name than `EnclosureComponent` try not to use `Component` in name to avoid collisions with view components

In regard of the form where it's used, we may call it `Slot`

PR #223, opened 2024-09-25
Update dependencies

nicolas-brousse:


vendor/javascript/html2pdf.js.bundle.js
B-Rass, line 1: PDF exports don't work anymore

B-Rass (changes requested):

B-Rass (approved):

PR #234, opened 2024-10-16
343808 - Add presence and format validation on Card name

nicolas-brousse:


spec/models/composant_spec.rb
```diff
@@ -17,6 +17,22 @@
 
   describe "validations" do
     xit { is_expected.to be_valid }
```
B-Rass, line 19: Did you look why this test is deactivated?

B-Rass (changes requested):
Maybe we should make it more explicit. Or make the field form red? Or both?

spec/models/composant_spec.rb
nicolas-brousse: The object need to have associations set correctly for that

nicolas-brousse:
> Maybe we should make it more explicit. Or make the field form red? Or both?

I improved the error message

B-Rass (approved):

PR #243, opened 2024-10-29
341621 Add attribute access control to Room, Islet and Bay

B-Rass:


B-Rass:
@nicolas-brousse I'm not totally sure about how I implement `access_control`

config/locales/fr.yml
```diff
@@ -683,6 +683,13 @@ fr:
       update: Modifier
       submit: Envoyer
 
+  concerns:
+    access_control:
+      badge: Badge
+      key: Clé
+      key_loken: Clé Loken
```
nicolas-brousse, line 690:
```suggestion
      locken_key: Clé Locken
```

Also rename other in the concern code

app/decorators/bay_decorator.rb
```diff
@@ -0,0 +1,15 @@
...
+    def access_control_options_for_select
+      Bay.access_controls.keys.map { |a_c| [I18n.t("concerns.access_control.#{a_c}"), a_c] }
+    end
+  end
+
+  def access_control_to_human
+    return I18n.t("concerns.access_control.blank") unless (a_c = access_control.presence)
+
+    I18n.t("concerns.access_control.#{a_c}")
+  end
```
nicolas-brousse, line 14: **Optional**. May be moved in concern to avoid repetition.

B-Rass: I was 50 pro and 50 cons adding it in concern rather than decorator.

I took this option because it's 100% decorator related code, and not "model"
Also, sure it will avoid repetition, but the code was less clear

If you have argument in favor of concern rather than decorator, I'd like to hear them.
I let you decide what you think is better

nicolas-brousse: Let stay like this but remove the `HasAccessControl` concern.

nicolas-brousse (approved):

B-Rass:
@nicolas-brousse conflicts solved

Issue #236, opened 2024-10-29
Better Localization sections


nicolas-brousse:
- On Room, move Localization information in a separate card
- On Frame, add more information about localization, like on Bays
- On AirConditioner, improve localization information
- On Server, move localization information in a separate card

PR #250, opened 2024-10-31
#236 - Better Localization sections 

B-Rass:


app/views/servers/show.html.erb
nicolas-brousse, line 1: Maybe we could move the location card on the right like on other show pages

B-Rass: done

nicolas-brousse (approved):

PR #252, opened 2024-11-06
Create Contacts

nicolas-brousse:
- [x] Add tests
- [x] Make Contact and ContactRole index columns sortables
- [x] Implement assignments, create CRUD and frame like for associations
- [x] I18n
- [x] List assignments on Contact show
- [x] In contact index regroup first_name and last_name

### DIFF
--- a/app/controllers/contact_roles_controller.rb
+++ b/app/controllers/contact_roles_controller.rb
@@ -6,8 +6,8 @@ class ContactRolesController < ApplicationController
   # GET /contact_roles
   # GET /contact_roles.json
   def index
-    # TODO: add processor for sort
-    @contact_roles = ContactRole.all
+    @filter = ProcessorFilter.new(ContactRole.all, params)
+    @contact_roles = @filter.results
   end
 
   # GET /contact_roles/1
--- a/app/controllers/contact_roles_controller.rb
+++ b/app/controllers/contact_roles_controller.rb
@@ -1,0 +1,81 @@
+# frozen_string_literal: true
+
+class ContactRolesController < ApplicationController
+  before_action :set_contact_role, only: %i[show edit update destroy]
+
+  # GET /contact_roles
+  # GET /contact_roles.json
+  def index
+    # TODO: add processor for sort
+    @contact_roles = ContactRole.all
+  end
+
+  # GET /contact_roles/1
+  # GET /contact_roles/1.json
+  def show; end
+
+  # GET /contact_roles/new
+  def new
+    @contact_role = ContactRole.new
+  end
+
+  # GET /contact_roles/1/edit
+  def edit; end
+
+  # POST /contact_roles
+  # POST /contact_roles.json
+  def create
+    @contact_role = ContactRole.new(contact_role_params)
+
+    respond_to do |format|
+      if @contact_role.save
+        format.html { redirect_to @contact_role, notice: t(".flashes.created") }
+        format.json { render :show, status: :created, location: @contact_role }
+      else
+        format.html { render :new }
+        format.json { render json: @contact_role.errors, status: :unprocessable_entity }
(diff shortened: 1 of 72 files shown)


B-Rass:
Ok I think I'm done with it. 

I thought it may be nice to always show contact assignment by default

nicolas-brousse:
> I thought it may be nice to always show contact assignment by default

Make sense, we have to move assignments index table in a partial and use it directly in the site show view. I'll look at it when I rework on this.

app/controllers/contact_roles_controller.rb
```diff
@@ -0,0 +1,81 @@
...
+        format.html { render :edit }
+        format.json { render json: @contact_role.errors, status: :unprocessable_entity }
+      end
+    end
+  end
+
+  # DELETE /contact_roles/1
+  # DELETE /contact_roles/1.json
+  def destroy
+    if @contact_role.destroy
```
B-Rass, line 58: I had a Rails error when I try to delete Role with an assignment.

Can't we just show a flash error message?

config/locales/fr.yml
```diff
@@ -653,7 +657,61 @@ fr:
...
+      flashes:
+        updated: Le role a bien été mis à jour.
+    destroy:
+      flashes:
+        destroyed: Le role a bien été supprimé.
+  contact_assignments:
+    index:
+      new_contact_assignment: Ajouter une assignation
+      title: Assignations de Contacts
+      delete_confirmation: Cette assignation ne pourra pas être supprimé tant qu'il est associé à au moins un Site. Voulez-vous continuer ?
```
B-Rass, line 701: I can delete with an assignation, so we should change the message here

test/fixtures/contact_assignments.yml
```diff
@@ -0,0 +1,11 @@
...
+
+# one:
+#   site: one
+#   contact: one
+#   contact_role: one
+
+# two:
+#   site: two
+#   contact: two
+#   contact_role: two
```
B-Rass, line 11: Why is it commented?

config/locales/fr.yml
```diff
@@ -653,7 +657,61 @@ fr:
         updated: La baie a bien été mise à jour.
     destroy:
       flashes:
-        destroyed: La baie bien été supprimé.
+        destroyed: La baie a bien étée supprimé.
+  contacts:
+    index:
+      new_contact: Ajouter un contact
+      title: Contacts
+      delete_confirmation: Ce contact ne pourra pas être supprimé tant qu'il est associé à au moins un Site. Voulez-vous continuer ?
```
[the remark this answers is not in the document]
nicolas-brousse: I don't think we should disallow user to remove a contact with assignation. But change the message to be coherent yes.

app/controllers/contact_roles_controller.rb
nicolas-brousse: I think there is a config missing in the model.

test/fixtures/contact_assignments.yml
nicolas-brousse: Not used and doesn't work if uncommented

B-Rass: ok

app/controllers/contact_roles_controller.rb
nicolas-brousse: fixed

config/locales/fr.yml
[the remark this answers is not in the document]
nicolas-brousse: fixed

config/locales/fr.yml
nicolas-brousse: fixed

B-Rass (approved):

PR #256, opened 2024-11-07
Add warning icon on bay if it has no frame associated

B-Rass:


app/components/page/heading_component.rb
```diff
@@ -16,7 +16,7 @@ class HeadingComponent < ApplicationComponent
           <% end %>
 
           <% if @title %>
-            <h1 class="text-center"><%= @title %></h1>
+            <h1 class="text-center"><%= @title.to_s.html_safe %></h1>
```
nicolas-brousse, line 19: Prefer to apply `html_safe` in the `show.html.erb` view. That way we apply it really when we want to. It's a bit more secure.

```erb
<%= render Page::HeadingShowComponent.new(resource: @bay, title: title.html_safe, breadcrumb_steps:) do |heading| %>
   ...
<% end %>
```

app/views/bays/index.html.erb
```diff
@@ -51,7 +51,8 @@
   <%= turbo_frame_tag(dom_id(Bay, :table), data: { turbo_action: :advance }) do %>
     <%= render List::DataTableComponent.new(@bays) do |table| %>
       <% table.with_column(Bay.human_attribute_name(:name)) do |bay| %>
-        <%= link_to bay, bay_path(bay), data: { turbo_frame: :_top } %>
+        <% link_text = bay.frames.any? ? bay.to_s : bay.decorated.no_frame_warning_icon %>
+        <%= link_to link_text, bay_path(bay), data: { turbo_frame: :_top } %>
```
nicolas-brousse, line 55: **Optional**. We may move the text in a block with condition.

nicolas-brousse (changes requested):

app/views/bays/index.html.erb
B-Rass: I don't get it

nicolas-brousse (approved):

PR #258, opened 2024-11-13
352641 add move cta on server show

B-Rass:


app/controllers/moves_controller.rb
```diff
@@ -13,7 +13,8 @@ def index
   def show; end
 
   def new
-    @move = Move.new(moveable_type: 'Server')
+    moveable_id = Server.friendly.select(:id).find(params[:server_id]).id if params[:server_id]
+    @move = Move.new(moveable_type: "Server", moveable_id: moveable_id)
```
nicolas-brousse, line 17:
```suggestion
    @move = Move.new(moveable_type: "Server", moveable_id: moveable_id)
    @move.moveable = Server.friendly.select(:id).find(params[:server_id]).id if params[:server_id].present?
```

Also maybe we should use `moveable_id` in query param?

nicolas-brousse (changes requested):
One small change suggestion, otherwise LGTM

nicolas-brousse (approved):

PR #259, opened 2024-11-15
353217 API: Fix search by name on server

B-Rass:


app/controllers/servers_controller.rb
```diff
@@ -6,6 +6,13 @@ class ServersController < ApplicationController
   before_action :set_server, only: [:show, :edit, :update, :destroy]
 
   def index
+    # Let server know that now name is not used anymore for research
```
nicolas-brousse, line 9:
```suggestion
    # Let server knows that now name is not used anymore for research
```

app/controllers/servers_controller.rb
```diff
@@ -6,13 +6,24 @@ class ServersController < ApplicationController
...
     @filter = ProcessorFilter.new(@servers, params)
 
-    @pagy, @servers = pagy(@filter.results)
-    @search_params = search_params
+    if request.format.json?
+      @servers = @filter.results
+    else
+      @pagy, @servers = pagy(@filter.results)
+      @search_params = search_params
+    end
```
nicolas-brousse, line 26: Suggestion

```suggestion
    @servers = @filter.results
    @search_params = search_params
    
    respond_to do |format|
      format.json
      format.html { @pagy, @servers = pagy(@server) }
    end
```

PR #273, opened 2024-11-27
Fix port type index

nicolas-brousse:
- Add filters on CardType
- Fix old link from PortType to CardType


B-Rass (approved):
It misses translation for the filter label. I added it

PR #282, opened 2024-12-04
Fix not found filter on external app records

B-Rass:


nicolas-brousse (changes requested):
Small typo, otherwise LGTM

PR #287, opened 2024-12-04
351548 - add filters on ExternalAppRecord sync with Category

nicolas-brousse:


app/controllers/external_app_records_controller.rb
```diff
@@ -6,7 +6,7 @@ def index
     @external_app_records_count = @external_app_records.count
     @filter = ProcessorFilter.new(@external_app_records, params)
     @external_app_records = @filter.results
-    @servers_count = Server.no_pdus.count
+    @servers_count = Server.glpi_synchronizable.count
```
B-Rass, line 9: To be removed (not used anymore)

app/views/categories/index.html.erb
```diff
@@ -27,6 +27,10 @@
       <% end %>
     <% end %>
 
+    <% table.with_column(Category.human_attribute_name(:is_glpi_synchronizable), sort_by: :is_glpi_synchronizable, class: "text-center", style: "min-width: 240px; width: 240px") do |category| %>
```
B-Rass, line 30:
```suggestion
    <% table.with_column(Category.human_attribute_name(:is_glpi_synchronizable), sort_by: :is_glpi_synchronizable, class: "text-center", style: "width: 240px") do |category| %>
```

B-Rass: `min-width` is not useful here

db/migrate/20241204143353_add_is_glpi_synchronizable_to_categories.rb
B-Rass, line 1: We talked about putting all the current categories to true. But we can let it that way

B-Rass (changes requested):

app/controllers/external_app_records_controller.rb
nicolas-brousse: Done

db/migrate/20241204143353_add_is_glpi_synchronizable_to_categories.rb
nicolas-brousse: I've memorized the opposite 😅

app/views/categories/index.html.erb
nicolas-brousse: Done

B-Rass (approved):

PR #298, opened 2024-12-11
343310 - Bulk action on Servers

MBausson:


MBausson:
About futures implementations of bulk actions on other resources:  

* The current `bulk_destroy` action could be moved to a concern  
* `delete :bulk_destroy` should be added to every resource's collection routes
* DataLists should specify the `bulk: true` constructor parameter
* Datalists should be wrapped within a `<form>` (see [this file](https://github.com/nanego/my-dcim/blob/b136623bc4088194e8d11638d265e67eba316728/app/views/servers/index.html.erb)). I'm thinking about moving this wrapper directly inside the DataList component

app/views/servers/index.html.erb
```diff
@@ -125,8 +125,18 @@
   <% end %>
 
   <%= turbo_frame_tag(dom_id(Server, :table), data: { turbo_action: :advance }) do %>
-    <div class="d-flex flex-column row-gap-4">
-      <%= render List::DataTableComponent.new(@servers) do |table| %>
+    <%= form_with url: bulk_destroy_servers_path, method: :delete, class: "d-flex flex-column row-gap-4", data: { controller: "bulk-actions" } do %>
+      <div style="visibility: hidden;" data-bulk-actions-target="actionsContainer">
+        <span class="fw-bolder" data-bulk-actions-target="checkedCount"></span> elements selected:
```
B-Rass, line 130: Localise the text

B-Rass (changes requested):
We should have ask user confirmation before deleting, with a text informing the user how many items is going to delete

nicolas-brousse:
We should have to thing a bit more of the future actions that may be implemented.
What if we have other actions, like edit or change server category?

I think at least the stimulus controller may be declared inside the datatable to make it easier than add it in every form.
Form may be inside datatable too but how to make it works with more than one action.

May be slot could be use to declare actions.

```rb
c.with_bulk_action("Destroy all", bulk_destroy_servers_path, method: :delete)
```

We also could imagine that bulk is automatically enabled when at least one "slot" is added.

config/routes.rb
```diff
@@ -55,6 +55,7 @@
       post :sort
       get :import_csv
       post :import
+      post :bulk_manage
```
nicolas-brousse, line 58: Prefer `patch`

app/controllers/concerns/bulk_actions.rb
```diff
@@ -0,0 +1,42 @@
...
+  included do
+    def bulk_manage
+      @items = params[:item_ids].map { |id| resource_class.find(id) }
+
+      if params[:bulk_destroy]
+        bulk_destroy
+      else
+        redirect_back fallback_location: root_path
+      end
+    end
```
nicolas-brousse, line 15: We may imagine having this just as a proxy, and having `bulk_destroy` in the ServerController.

And if we have `bulk_disable` button later, then we may just need to have a `bulk` disable action in the controller.

app/components/list/data_table_component.rb
```diff
@@ -20,25 +21,62 @@ def call
...
     end
 
-    private
+    def render_bulk_head_cell
+      render(List::TableComponent::TableHeadCell.new(style: "width: 0;")) do
+        tag.input class: "form-check-input", type: "checkbox", id: "checkboxAll", data: { bulk_actions_target: "checkboxAll" }
+      end
+    end
+
+    def render_bulk(&)
```
nicolas-brousse, line 62:
```suggestion
    def render_bulk
```

nicolas-brousse (changes requested):

nicolas-brousse (approved):

B-Rass (approved):

PR #304, opened 2024-12-18
Extract PDUs from Servers views

B-Rass:


spec/requests/servers_request_spec.rb
```diff
@@ -5,12 +5,14 @@
...
   let(:server) { servers(:one) }
   let(:server2) { servers(:two) }
+  let(:pdu) { servers(:pdu) }
 
   before do
     sign_in users(:one)
 
     server.save!
     server2.save!
+    pdu.save!
```
nicolas-brousse, line 15: Why do we need to save the record that came from fixtures? There are not already saved in database?

app/views/servers/_form.html.erb
```diff
@@ -23,7 +23,7 @@
       <fieldset class="col-md-6 col-12">
         <%= f.label :modele_id, class: "form-label" %>
         <%= f.collection_select(:modele_id,
-                                Modele.all_sorted,
+                                Modele.no_pdus.sorted,
```
nicolas-brousse, line 26: Don't we need to keep use `all_sorted` scope?

app/controllers/servers_controller.rb
```diff
@@ -127,8 +127,8 @@ def destroy_connections
 
   # Use callbacks to share common setup or constraints between actions.
   def set_server
-    @server = Server.where('lower(numero) = ?', params[:id].to_s.downcase).includes(:cards => [:ports => %i[connection cable], :card_type => [:port_type]]).first
-    @server = Server.friendly.find(params[:id].to_s.downcase) unless @server
+    @server = Server.includes(cards: [ports: %i[connection cable], card_type: [:port_type]])
+      .find_with_numero_or_friendly_id(params[:id].to_s.downcase)
```
nicolas-brousse, line 131: Not sure we need to keep the `include` since we aren't in a collection.

app/views/servers/show.html.erb
```diff
@@ -105,15 +103,19 @@
           <dl class="show-page_dl d-grid row-gap-2 mb-0">
             <dt class="pb-2"><%= Modele.model_name.human %></dt>
             <dd class="mb-0 pb-2 ps-3">
-              <%= link_to @server.modele, modele_path(@server.modele) if @server.modele.try(:name) %>
+              <% if @server.modele&.architecture.present? %>
+                <%= link_to @server.modele.architecture, architecture_path(@server.modele.architecture) %>
+              <% end %>
```
nicolas-brousse, line 108: Is this a mistake? Model has been replaced by architecture

app/views/servers/_form.html.erb
B-Rass: I don't see why

app/views/servers/show.html.erb
B-Rass: I fucked up my rebase

app/views/servers/_form.html.erb
nicolas-brousse: Because `all.sorted` and `all_sorted` doesn't make the same sort

B-Rass: updated

app/controllers/power_distribution_units_controller.rb
```diff
@@ -0,0 +1,106 @@
...
+      flash[:alert] = t(".flashes.connections_not_destroyed")
+    end
+
+    redirect_to power_distribution_unit_path(@pdu)
+  end
+
+  private
+
+  def set_pdu
+    @pdu = Server.friendly_find_by_numero_or_name(params[:id].to_s.downcase)
```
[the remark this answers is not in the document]
B-Rass: yes stupid me

nicolas-brousse (approved):

PR #320, opened 2025-01-30
363399 - Fixing port position not displaying when 1FC

B-Rass:


app/helpers/servers_helper.rb
```diff
@@ -119,16 +119,19 @@ def link_to_port_without_label(position, port_data, port_type, card_id, port_id)
...
   def link_to_port(position, port_data, port_type, card_id, port_id, default_label = '')
-    cable_name = port_data.try(:cable_name).present? ? port_data.try(:cable_name) : default_label
+    cable_name = port_data.try(:cable_name).present? ? port_data.cable_name : default_label
+
     case port_type.name
     when 'RJ', 'XRJ'
       port_type_name = "RJ"
     when 'FC', 'SC'
       port_type_name = "FC"
+      cable_name = "0#{position}" if cable_name.blank?
```
nicolas-brousse, line 129: What the prefixed `0` is for?

app/helpers/servers_helper.rb
```diff
@@ -119,16 +119,19 @@ def link_to_port_without_label(position, port_data, port_type, card_id, port_id)
   end
 
   def link_to_port(position, port_data, port_type, card_id, port_id, default_label = '')
-    cable_name = port_data.try(:cable_name).present? ? port_data.try(:cable_name) : default_label
+    cable_name = port_data.try(:cable_name).present? ? port_data.cable_name : default_label
```
nicolas-brousse, line 122:
```suggestion
    cable_name = port_data&.cable_name.presence || default_label
```

app/helpers/servers_helper.rb
B-Rass: to display `01` and not `1`

nanego: You could use position.to_s.rjust(2, "0") to ensure the number is displayed with at least two digits.

nicolas-brousse: There is also `format("%02d", position)` method that could be use

nicolas-brousse (approved):

PR #325, opened 2025-02-05
366659 - Add server type filter

MBausson:


app/views/servers/index.html.erb
B-Rass, line 1: We should rename the `type` column to be consistent with the filter and the rest of the application.

So, add update the `Type` column to use `modele.category` instead.

(remove the unused locale + check everywhere that is not use anymore)

B-Rass (approved):

PR #326, opened 2025-02-05
367616 - Add overview block to site room and islet

B-Rass:


app/views/rooms/show.html.erb
```diff
@@ -90,6 +90,32 @@
...
+                <h6 class="mb-2">
+                  <%= render LabelComponent.new(type: :primary) do %>
+                    <%= link_to("Ilot #{islet}", visualization_room_path(@room, islet: islet.name, view: params[:view], bg: params[:bg])) %>
+                  <% end %>
+                </h6>
+              <% end %>
+              <%= render "bays/overview", bays: islet.decorated.overviewed_bays_array, islet: islet.name %>
+            </div>
+          <% end %>
+        <% end %>
```
nicolas-brousse, line 114: @B-Rass maybe this part could be that `rooms/_room_overview.html.erb` contains to be reused in rooms#show and rooms overview.

Seems to be the same visual structure.

B-Rass: Yes good call, I'm going to update it right now

B-Rass: Finally no

PR #328, opened 2025-02-05
358734 - Add filters on GLPI synchronization index

MBausson:


MBausson:
The textual filter for servers' names does not work, this is the sql statement and the error thrown:
```sql
SELECT COUNT(*) FROM "external_app_records" INNER JOIN "servers" "server" ON "server"."id" = "external_app_records"."server_id" WHERE "servers"."name" ILIKE '%A%'
```

```ruby
ActiveRecord::StatementInvalid (PG::UndefinedTable: ERROR:  invalid reference to FROM-clause entry for table "servers" LINE 1: ..."."id" = "external_app_records"."server_id" WHERE "servers"....
```

app/processors/external_app_records_processor.rb
```diff
@@ -3,13 +3,21 @@
...
   map :frame_ids, filter_with: :non_empty_array do |frame_ids:|
     raw.joins(:frame).where(frame: { id: frame_ids })
   end
 
+  map :modele_ids, filter_with: :non_empty_array do |modele_ids:|
+    raw.joins(:server).where(servers: { modele_id: modele_ids })
+  end
+
+  map :server_name do |server_name:|
+    raw.joins(:server).where(Server.arel_table[:name].matches("%#{server_name}%"))
```
nicolas-brousse, line 18:
```suggestion
    raw.joins(:server).where(Server.arel_table[:name].alias(:server).matches("%#{server_name}%"))
```

Try this to fix the issue

app/controllers/external_app_records_controller.rb
```diff
@@ -33,6 +33,6 @@ def sync_all_servers_with_glpi
   end
 
   def settings_params
-    params[:external_app_record_setting]&.permit(category_ids: []) || {}
+    params[:external_app_record_setting]&.permit(category_ids: [], modele_ids: []) || {}
```
nicolas-brousse, line 36: Not sure `modele_ids` is required here

app/processors/external_app_records_processor.rb
MBausson: It worked, the alias must be set before the `[:name]`.  However, I had to modify the Modele filter to use `server` in the query.  
```ruby
map :modele_ids, filter_with: :non_empty_array do |modele_ids:|
  raw.joins(:server).where(server: { modele_id: modele_ids })
end
```

MBausson: Now we have the same issue but on the tests (spec/processors/external_app_records_processor_spec.rb)... I've been working on a fix, but I'm not sure it is the right way of doin it

spec/processors/external_app_records_processor_spec.rb
B-Rass, line 1: Update `describe "When searching on every fields" do` test

B-Rass (changes requested):

B-Rass (approved):

PR #347, opened 2025-02-26
342607 - Remove Moves#show.html action

B-Rass:


config/routes.rb
```diff
@@ -16,6 +17,8 @@
       match 'update_connection', to: 'moves#update_connection', via: %i[patch post put]
     end
   end
+  resources :moves, only: %i[show], constraints: ->(request) { request.format == :json }
```
nicolas-brousse, line 20: Why not keeping it in the initial definition? Since specify the format with `respond_to`

B-Rass: To remove also the route, but maybe it's not useful. I dunno

nicolas-brousse: To have a `404`?

B-Rass: Yah

PR #350, opened 2025-02-27
309 - In Enclosure make display vertical by default

B-Rass:


app/models/enclosure.rb
```diff
@@ -1,14 +1,14 @@
...
 class Enclosure < ApplicationRecord
-  DISPLAYS = %i[vertical horizontal grid].freeze
-
   has_changelog
   acts_as_list scope: [:modele_id]
 
   belongs_to :modele
   has_many :composants, dependent: :destroy
 
+  enum :display, { vertical: 0, horizontal: 1, grid: 2 }, default: :vertical, validate: true
```
nicolas-brousse, line 10: A special though to not to have keep value as string? Like for `AirConditioner#status`

B-Rass: I mimic what I have done on `Room#status`, and `integer` is better than `string` no?

I have a remember to have read that, that it's better to user enum with integers

That's why

PR #353, opened 2025-03-06
368436 - Clean Server model from unused and old attributes

B-Rass:
- [x] Remove attributes
- [x] Execute migration
- [x] Remove attributes used by JS
- [x] Test pages


db/migrate/20250227170947_remove_not_used_attributes_from_server.rb
```diff
@@ -0,0 +1,16 @@
...
+    change_table :servers, bulk: true do |t|
+      t.remove :pdu_ondule, :pdu_normal, type: :string
+      t.remove :fc_total, :fc_utilise, :fc_calcule, :fc_futur, type: :integer
+      t.remove :rj45_total, :rj45_utilise, :rj45_cm, :rj45_calcule, :rj45_futur, type: :integer
+      t.remove :ipmi_utilise, :ipmi_futur, :ipmi_dedie, type: :integer
+      t.remove :tenGbps_futur, type: :integer
+      t.remove :hostname, type: :string
+      t.remove :etat_conf_reseau, :action_conf_reseau, type: :string
+      t.remove :i, :ip, type: :string
+    end
```
nicolas-brousse, line 14: Could be easier to use a revert here to make it more readable?

nicolas-brousse (commented):
One suggestion otherwise LGTM

db/migrate/20250227170947_remove_not_used_attributes_from_server.rb
B-Rass: I'm not convinced by the side more readable by using a `revert`

nicolas-brousse: okay let's keep it like this. Just to let you know, for my perspective, it's because with revert, we got the column name directly after `t.*` instead at the end. But has you prefer, it's not really important

B-Rass: I prefer it like this, to see the attribute name first

nicolas-brousse (approved):

PR #354, opened 2025-03-06
342607 - Remove ServerState model

B-Rass:


db/migrate/20250306110910_remove_server_states.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+
+class RemoveServerStates < ActiveRecord::Migration[8.0]
+  def change
+    revert do
+      create_table :server_states do |t|
+        t.string :name
+
+        t.timestamps null: false
```
nicolas-brousse, line 9:
```suggestion
        t.timestamps
```

I think it's the defaults

nicolas-brousse (changes requested):
Also maybe we needs to clean the API documentation too

B-Rass:
> Also maybe we needs to clean the API documentation too

done

nicolas-brousse (approved):

PR #380, opened 2025-04-17
367 improve breadcrumb new system

B-Rass:
- [Rename add to add_step](https://github.com/nanego/my-dcim/commit/1b4ec178c38f2b61618046ea07da81d961096ae2)
- [I18n on missing strings](https://github.com/nanego/my-dcim/commit/3a0ecb34524b651621205a25987fb45dfb357cb0)

### DIFF
--- a/app/controllers/bays_controller.rb
+++ b/app/controllers/bays_controller.rb
@@ -10,6 +10,9 @@ class BaysController < ApplicationController
   columns_preferences_with model: Bay, default: DEFAULT_COLUMNS, available: AVAILABLE_COLUMNS
 
   before_action :set_bay, only: %i[edit update destroy show]
+  before_action except: %i[index] do
+    breadcrumb.add_step(t("bays.index.title"), bays_path)
+  end
 
   def index
     @bays = Bay.joins(islet: :room).order("rooms.position, islets.name, bays.lane, bays.position")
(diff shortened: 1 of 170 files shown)


app/controllers/bays_controller.rb
```diff
@@ -4,6 +4,9 @@ class BaysController < ApplicationController
   include RoomsHelper
 
   before_action :set_bay, only: %i[edit update destroy show]
+  before_action except: %i[index] do
+    breadcrumb.add_step(t("bays.index.title"), bays_path)
```
nicolas-brousse, line 8:
```suggestion
    breadcrumb.add_step(t(".index.title"), bays_path)
```

does this works?

app/controllers/power_distribution_units_controller.rb
```diff
@@ -2,6 +2,10 @@
 
 class PowerDistributionUnitsController < ApplicationController
   before_action :set_pdu, only: %i[show edit update destroy destroy_connections]
+  before_action except: %i[index] do
+    # TODO: prefer use ActiveRecord translation of Modele name
+    breadcrumb.add_step(t("power_distribution_units.index.title"), power_distribution_units_path)
```
nicolas-brousse, line 7: This todo should stay or be applied yet?

nicolas-brousse (changes requested):

app/controllers/bays_controller.rb
B-Rass: Because it's just `Bay`, we prefer `Bay and double bays` here

app/controllers/power_distribution_units_controller.rb
B-Rass: it should stay, PDU is not a modele yet

nicolas-brousse (approved):

PR #398, opened 2025-05-14
Add an option to "create and create another one"

### LINKED ISSUE #332: Add an option to "create and create another one"
Two options : 
- With a button
- With a checkbox

Image

MBausson:
Closes #332

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -49,9 +49,7 @@ class ApplicationController < ActionController::Base
   helper_method :breadcrumb
 
   def redirect_to_new_or_to(options = {}, response_options = {})
-    if params[:create_another_one].present?
-      options = url_for(action: :new)
-    end
+    options = url_for(action: :new) if params[:create_another_one].present?
 
     redirect_to options, response_options
   end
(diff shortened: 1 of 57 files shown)


MBausson:
I added the shared_example on controllers that have a rspec request spec, and not on others

app/controllers/application_controller.rb
```diff
@@ -48,6 +48,14 @@ def breadcrumb
...
   helper_method :breadcrumb
 
+  def create_another_one_path(default_path)
+    if params[:create_another_one].present?
+      url_for(action: :new) || root_path
+    else
+      default_path
+    end
+  end
+
```
nicolas-brousse, line 58: I suggest to have a custom `redirect_to` helper like `redirect_to_new_or_to air_conditioner_path(@air_conditioner)`


```rb
  def redirect_to_new_or_to(options = {}, response_options = {})
    if params[:create_another_one].present?
      options = url_for(action: :new)
    end

    redirect_to options, response_options
  end
```

Something like that

nicolas-brousse (commented):
One suggestion otherwise LGTM

MBausson:
> I think the checkbox should stay checked after a create

Done, I've made so the `create_another_one` url param stays in the redirected URL when creating another one

B-Rass (approved):

PR #408, opened 2025-05-28
Fix 'U' column on Servers index

MBausson:
The column should display the amount of U being used, and not the total number of U


app/controllers/servers_controller.rb
B-Rass, line 1: Why `u_used` and not just `u`? 

If possible, I think it's better to keep consistency with `modele`.

B-Rass (approved):

PR #424, opened 2025-06-13
383737 - Add prev_position to Moves

B-Rass:
- [x] Create `prev_position` for `Moves`
- [x] Make a migration for existing data
- [x] Use it on views (table)
- [x] Use `prev_frame` instead of `moveable.frame`
- [x] Add tests

### DIFF
--- a/app/controllers/moves_controller.rb
+++ b/app/controllers/moves_controller.rb
@@ -33,6 +33,7 @@ class MovesController < ApplicationController # rubocop:disable Metrics/ClassLen
     @move = @moves_project_step.moves.build(move_params)
 
     @move.prev_frame_id = @move.moveable.try(:frame_id)
+    @move.prev_position = @move.moveable.try(:position)
 
     if params[:move][:remove_connections] == 'Oui'
       @move.clear_connections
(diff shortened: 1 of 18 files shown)


app/models/move.rb
```diff
@@ -42,6 +42,14 @@ def execute!(apply_connections: true)
       if equipment.save!
         MovedConnection.per_servers([equipment]).map(&:execute!) if apply_connections
 
+        # Update prev_frame and prev_position for incoming moves
+        Move.not_executed
+          .where(moveable: equipment)
+          .where.not(id: self)
+          .find_each do |move|
+            move.update(prev_frame_id: frame.id, prev_position: position)
+          end
```
nicolas-brousse, line 51: We also may update `prev_frame` of current move via moveable I think

app/controllers/moves_controller.rb
```diff
@@ -179,7 +178,7 @@ def set_move
 
   # Never trust parameters from the scary internet, only allow the white list through.
   def move_params
-    params.expect(move: %i[moveable_type moveable_id frame_id position prev_frame_id])
+    params.expect(move: %i[moveable_type moveable_id frame_id position])
```
nicolas-brousse, line 181: This was not used already?

nicolas-brousse (changes requested):

PR #428, opened 2025-06-19
383737 - Improve moves frames visualisation

nicolas-brousse:
- [x] Tests
- [x] upload frame view on move creation / edition


app/controllers/moves_project_steps_controller.rb
```diff
@@ -39,17 +39,7 @@ def set_moves_project_step
 
   def set_frame_updated
     @frame = Frame.friendly.find(params[:frame_id])
-
-    @moves = @moves_project_step.moves.where(frame: @frame, moveable_type: "Server")
```
B-Rass, line 43: It's used in `app/views/moves_project_steps/frame.html.erb`, to show the moves applied

B-Rass (approved):

PR #430, opened 2025-06-25
390895 - Admin can update user information

MBausson:
They should be able to edit
- Email
- Name

Use policies to authorize access to such action

### DIFF
--- a/spec/requests/users_request_spec.rb
+++ b/spec/requests/users_request_spec.rb
@@ -26,7 +26,8 @@ RSpec.describe "Users" do
     context "with regular user" do
       include_context "with authenticated user"
 
-      it { expect { response }.to raise_error(ActionPolicy::Unauthorized) }
+      it { expect(response).to have_http_status(:redirect) }
+      it { expect(response).to redirect_to(root_path) }
     end
   end
 
@@ -60,7 +61,8 @@ RSpec.describe "Users" do
     context "with regular user" do
       include_context "with authenticated user"
 
-      it { expect { response }.to raise_error(ActionPolicy::Unauthorized) }
+      it { expect(response).to have_http_status(:redirect) }
+      it { expect(response).to redirect_to(root_path) }
     end
 
     context "with user is current_user" do
@@ -92,7 +94,8 @@ RSpec.describe "Users" do
     context "with regular user" do
       include_context "with authenticated user"
 
-      it { expect { response }.to raise_error(ActionPolicy::Unauthorized) }
+      it { expect(response).to have_http_status(:redirect) }
+      it { expect(response).to redirect_to(root_path) }
     end
   end
 
@@ -157,7 +160,8 @@ RSpec.describe "Users" do
         let(:user) { users(:one) }
       end
 
-      it { expect { response }.to raise_error(ActionPolicy::Unauthorized) }
+      it { expect(response).to have_http_status(:redirect) }
+      it { expect(response).to redirect_to(root_path) }
     end
   end
 
@@ -183,7 +187,8 @@ RSpec.describe "Users" do
         let(:user) { users(:one) }
       end
 
-      it { expect { response }.to raise_error(ActionPolicy::Unauthorized) }
+      it { expect(response).to have_http_status(:redirect) }
+      it { expect(response).to redirect_to(root_path) }
     end
 
     context "with valid parameters" do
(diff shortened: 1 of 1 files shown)


app/controllers/users/registrations_controller.rb
```diff
@@ -2,8 +2,18 @@
...
-    before_action :admin_only, only: [:new, :create]
+    def edit_user; end
+
+    def update_user
+      if @user.update(user_params)
+        redirect_to user_path(@user), notice: t(".flashes.updated")
+      else
+        render :edit_user, status: :unprocessable_entity
+      end
+    end
```
B-Rass, line 16: Use classic `edit` and `update` in `UsersController` instead

app/views/users/show.html.erb
```diff
@@ -4,7 +4,18 @@
   breadcrumb.add_step(title)
 %>
 
-<%= render Page::HeadingComponent.new(title:, breadcrumb:, back_button_url: users_path) %>
+<%= render Page::HeadingComponent.new(title:, breadcrumb:, back_button_url: users_path) do |heading| %>
```
B-Rass, line 7: We may be able to use the `HeadingShowComponent` here

app/views/users/index.html.erb
```diff
@@ -90,6 +90,14 @@
               <span class="visually-hidden"><%= t("action.suspend") %></span>
             <% end %>
           <% end %>
+          <% if allowed_to?(:edit?, user) %>
+            <%= link_to edit_user_path(user), class: "btn btn-outline-primary" do %>
```
B-Rass, line 94:
```suggestion
            <%= link_to edit_user_path(user), class: "btn btn-outline-info" do %>
```

app/views/users/index.html.erb
```diff
@@ -90,6 +90,14 @@
...
+              <span class="bi bi-pencil" title="<%= t("action.edit") %>" aria-hidden="true"
+                                                                          data-controller="tooltip"
+                                                                          data-bs-placement="left"></span>
+              <span class="visually-hidden"><%= t("action.edit") %></span>
+            <% end %>
+          <% end %>
           <%= link_to user_path(user), method: :delete, data: { confirm: t("action.confirm") },
                                                         class: "btn btn-danger" do %>
```
B-Rass, line 102:
```suggestion
                                                        class: "btn btn-outline-danger" do %>
```

app/views/users/edit.html.erb
```diff
@@ -0,0 +1,28 @@
+<%
+  breadcrumb.add_step(@user.name, user_path(@user))
+  breadcrumb.add_step(title = t(".title"))
+%>
+
+<%= render Page::HeadingComponent.new(title: title, breadcrumb:) %>
+
+<div class="col-12 p-4 border-top">
+  <%= form_with model: @user, url: user_path(@user), method: :patch, class: "col-12 col-md-10 col-lg-8 mx-auto" do |form| %>
+    <%= render FormErrorsComponent.new(@user) %>
```
B-Rass, line 10: It doesn't show up when I have error on my form

app/views/users/show.html.erb
```diff
@@ -4,7 +4,8 @@
   breadcrumb.add_step(title)
 %>
 
-<%= render Page::HeadingComponent.new(title:, breadcrumb:, back_button_url: users_path) %>
+<%= render Page::HeadingShowComponent.new(resource: @user, title:, breadcrumb:) %>
```
B-Rass, line 7: The `back_button` doesn't appear

spec/requests/users_request_spec.rb
```diff
@@ -134,23 +136,56 @@
...
-      patch user_path(record), params: params
+      patch(user_path(record), params: params)
 
       # NOTE: used to simplify usage and custom test done in final spec file.
       @response # rubocop:disable RSpec/InstanceVariable
     end
 
     let(:record) { User.create!(email: "user-target@example.com", password: "passwordpassword") }
-    let(:valid_attributes) { { role: "admin" } }
+    let(:valid_attributes) { { role: "admin", name: "newFirstName" } }
```
B-Rass, line 173: Why did you add `name` here and did not use it?

B-Rass (changes requested):
You have to manage error when user try to access pages he has no right : 

image

B-Rass (changes requested):
Also prevent an admin to delete and lock/unlock his own account, via the policy. And use the policy to know if we should display the button or not.

B-Rass (changes requested):
Also we can add the `name` column now in the index table, before `email`

B-Rass (approved):

PR #432, opened 2025-06-25
Improve refresh_data with previous step data

B-Rass:


nicolas-brousse (commented):
One small suggestion otherwise LGTM

PR #434, opened 2025-06-26
372044 - Add a shortcut to add /remove frame and bay on Room#overview

B-Rass:
- [x] Manage redirection when overview view is used on show pages.
- [x] Remove `acts_as_list` on frames
- [x] Remove `acts_as_list` on bays


B-Rass:
Manage redirection when overview view is used on show pages.

app/assets/stylesheets/application.scss
```diff
@@ -22,8 +22,6 @@
  *= require tom-select.bootstrap5
  */
 
-@import "jquery-ui.css";
```
nicolas-brousse, line 25: Are we sure we don't need it?

app/controllers/application_controller.rb
```diff
@@ -65,6 +65,16 @@ def form_redirect_to(options = {}, response_options = {})
     end
   end
 
+  def form_redirect_to_new_or_to(options = {}, response_options = {})
+    if params[:create_another_one].present?
+      options = url_for(action: :new, create_another_one: "1", redirect_to_on_success: params[:redirect_to_on_success])
```
nicolas-brousse, line 70: Not sure `redirect_to_on_success` is usefull here. We can't really do both `redirect_to_new_or_to` and `form_redirect_to` at the same time.

app/controllers/application_controller.rb
nicolas-brousse, line 61:
```suggestion
    elsif params[:redirect_to_on_success].present?
```

app/views/bays/_form.html.erb
```diff
@@ -1,4 +1,6 @@
 <%= form_for(@bay, html: { class: "col-12 col-md-10 col-lg-8 mx-auto", role: "form" }) do |f| %>
+  <%= hidden_field_tag(:redirect_to_on_success, params[:redirect_to_on_success]) if params[:redirect_to_on_success] %>
```
nicolas-brousse, line 2: Maybe we could have an helper here `<%= redirect_to_on_success_hidden_field_tag(params[:redirect_to_on_success]) %>` that do the job for us. And use it in other places

app/assets/stylesheets/application.scss
B-Rass: It's already called in the line above. 

I'm not totally sure where it's used though, but for me it's use on drag'n'dropping server on bay visualization. And I didn't see regression while I was testing it

app/views/bays/_form.html.erb
B-Rass: Sure why not. Where to add this helper?

nicolas-brousse: in a `FormHelper` I guess

PR #438, opened 2025-07-03
Add select on pagination to allow user choose number of items per page

### LINKED ISSUE #297: Add select on pagination to allow user choose number of items per page
On paginated index, add a select `[25, 50, 100, 150, 200]` to let user choose the number of items we want to be shown on a single page.

In the same time, move the pagination in a Component.

MBausson:
Closes #297


MBausson:
TODO: Make "Items per page" a user settings, default set to 100

app/components/pagination_component.rb
```diff
@@ -0,0 +1,52 @@
+# frozen_string_literal: true
+
+class PaginationComponent < ApplicationComponent
+  include Pagy::Frontend
+
+  DEFAULT_ITEMS_PER_PAGE = 100
+
+  erb_template <<~ERB
+    <div class="pagination-component">
```
B-Rass, line 9: Add the CSS rules you define in `application.scss` here

config/locales/activerecord.en.yml
```diff
@@ -113,6 +113,8 @@ en:
         bay_id: Bay
         modele_category_id: Category
       theme: Theme
+      user:
+        items_per_page: Items per page
```
B-Rass, line 117: Could be nice to add the other attributes here (those on the settings page)

B-Rass (changes requested):

MBausson:
> Don't we want to always display the `items_per_page select`, even if there is only one page?
> 
> What do you guys @nicolas-brousse @MBausson think about it?

Yes why not, I don't see reasons not to show it

app/components/pagination_component.rb
```diff
@@ -0,0 +1,52 @@
+# frozen_string_literal: true
+
+class PaginationComponent < ApplicationComponent
+  include Pagy::Frontend
+
+  DEFAULT_ITEMS_PER_PAGE = 100
+
+  erb_template <<~ERB
+    <div class="d-inline-flex align-items-baseline gap-2 flex-wrap" style="width: fit-content;">
```
B-Rass, line 9: `style="width: fit-content;"`

It would be cleaner to add it here: `app/assets/stylesheets/components/pagination_component.scss`

B-Rass: Sorry I wasn't clean enough on my last review ..

B-Rass (changes requested):

app/controllers/external_app_records_controller.rb
```diff
@@ -6,13 +6,13 @@ class ExternalAppRecordsController < ApplicationController
   end
 
   def index
-    @external_app_records = ExternalAppRecord.includes(server: :frame).order("servers.name")
+    @external_app_records = ExternalAppRecord.includes(:server, :frame).order("servers.name")
```
nicolas-brousse, line 9: Why this change?

app/components/pagination_component.rb
```diff
@@ -0,0 +1,52 @@
+# frozen_string_literal: true
+
+class PaginationComponent < ApplicationComponent
+  include Pagy::Frontend
+
+  DEFAULT_ITEMS_PER_PAGE = 100
```
nicolas-brousse, line 6: This may be moved in User model I think

app/models/user.rb
```diff
@@ -20,12 +21,14 @@ class User < ApplicationRecord
   store_attribute :settings, :theme, :string, default: AVAILABLE_THEMES.first
   store_attribute :settings, :visualization_bay_default_background_color, :string, default: AVAILABLE_BAY_BACKGROUND_COLORS.first
   store_attribute :settings, :visualization_bay_default_orientation, :string, default: AVAILABLE_BAY_ORIENTATIONS.first
+  store_attribute :settings, :items_per_page, :integer, default: 100
```
nicolas-brousse, line 24: Use a constant for the default

config/initializers/pagy.rb
```diff
@@ -213,7 +213,6 @@
 require "pagy/extras/size"
 Pagy::DEFAULT[:size] = [2, 2, 2, 2]
 
-require "pagy/extras/limit"
```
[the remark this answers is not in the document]
nicolas-brousse: Make as it was. I prefer to have all config at the bottom to make it easier to read

app/components/pagination_component.rb
```diff
@@ -0,0 +1,52 @@
...
+            :items_per_page,
+            options_for_select(options, selected_items_per_page),
+            onchange: "window.location.href = this.value",
+            class: "form-select form-select-sm"
+          ) %>
+      </div>
+    </div>
+  ERB
+
+  def initialize(pagy:, params:, default_limit:)
```
nicolas-brousse, line 24: Why not using `limit` method from controller, to give the final value, instead having `params` and `default_limit`?

nicolas-brousse (changes requested):

app/controllers/external_app_records_controller.rb
MBausson: This fixes a N+1 query on ExternalAppRecords#Index. In this view, we get frames via `record.frame`, and not `record.server.frame`

nicolas-brousse (changes requested):

nicolas-brousse (approved):

nicolas-brousse (approved):

PR #440, opened 2025-07-09
Add a create button on HasManyFrame component

### LINKED ISSUE #396: Add a create button on HasManyFrame component
The create button should pass params to pre-filled data from where the button has been hit.

Example : hitting the create a frame in a bay, make the form pre-filled with bay

MBausson:
Closes #396

### DIFF
--- a/app/components/has_many_turbo_frame_component.rb
+++ b/app/components/has_many_turbo_frame_component.rb
@@ -14,22 +14,22 @@ class HasManyTurboFrameComponent < ApplicationComponent
           <% if actions? %>
             <%= actions %>
           <% else %>
-          <div>
+          <span>
+            <%= link_to @url, class: class_names("link-light", "link-dark": @type == :warning) do %>
+              <%= t(".see_more_filters") %>
+              <% end %>
+            <% end %>
+
             <% if @new_path %>
-              <%= render ButtonComponent.new(t("action.create"),
+              <%= render ButtonComponent.new(@new_label,
                                             url: @new_path,
                                             icon: "plus-lg",
                                             size: :sm,
                                             variant: :success,
                                             is_responsive: true,
-                                            extra_classes: "me-3") %>
+                                            extra_classes: "ms-3") %>
             <% end %>
-
-            <%= link_to @url, class: class_names("link-light", "link-dark": @type == :warning) do %>
-              <%= t(".see_more_filters") %>
-              <% end %>
-            <% end %>
-          </div>
+          </span>
         </div>
       <% end %>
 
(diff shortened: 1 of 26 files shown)


app/controllers/modeles_controller.rb
```diff
@@ -21,7 +21,7 @@ def show
   end
 
   def new
-    @modele = Modele.new
+    @modele = Modele.new(new_modele_params)
```
nicolas-brousse, line 24:
```suggestion
    @modele = Modele.new(modele_params)
```

Why not using existing `modele_params`?

app/components/has_many_turbo_frame_component.rb
```diff
@@ -14,10 +14,23 @@ class HasManyTurboFrameComponent < ApplicationComponent
           <% if actions? %>
             <%= actions %>
           <% else %>
-        <%= link_to @url, class: class_names("link-light", "link-dark": @type == :warning) do %>
+
+          <div>
```
nicolas-brousse, line 18: What this new div is for?

MBausson: The container is using `justify-content-between`, if I added the new create button outside of this div, it would have been centered to the middle

app/controllers/modeles_controller.rb
MBausson: `modele_params` expects a `modele` first, which makes giving parameters to path helpers less "clean", for example:  
`new_modele_path("modele[architecture_id]" => @architecture.id)`

I changed it and we are now using `modele_params`

app/components/has_many_turbo_frame_component.rb
B-Rass: Prefer the use of `span`

app/components/has_many_turbo_frame_component.rb
```diff
@@ -14,10 +14,22 @@ class HasManyTurboFrameComponent < ApplicationComponent
           <% if actions? %>
             <%= actions %>
           <% else %>
-        <%= link_to @url, class: class_names("link-light", "link-dark": @type == :warning) do %>
+          <div>
+            <% if @new_path %>
```
B-Rass, line 18: Add button after the `see more filters` link

app/components/has_many_turbo_frame_component.rb
```diff
@@ -14,10 +14,22 @@ class HasManyTurboFrameComponent < ApplicationComponent
...
+              <%= render ButtonComponent.new(t("action.create"),
```
B-Rass, line 19: I'm thinking maybe using a different wording here would be better: `Ajouter` instead of `Créer`

Maybe something dynamic like `Ajouter un modèle`

B-Rass (changes requested):

PR #446, opened 2025-07-16
Add unique constraint to prevent a moveable to be moved twice in a step

B-Rass:


app/models/move.rb
```diff
@@ -10,6 +10,8 @@ class Move < ApplicationRecord
   belongs_to :frame
   belongs_to :prev_frame, class_name: "Frame"
 
+  validates :moveable_id, uniqueness: { scope: [:step, :moveable_type] }
```
nicolas-brousse, line 13:
```suggestion
  validates :moveable, uniqueness: { scope: [:step] }
```

This doesn't work?

nicolas-brousse (commented):
One question otherwise LGTM

app/models/move.rb
B-Rass: Nope

B-Rass: I can't remember what I've done yesterday ... so I'm going to double check. I let you know

B-Rass: I have this error if I do it like you suggest

B-Rass:
```
ArgumentError in MovesController#create
Polymorphic associations do not support computing the class.
```

PR #449, opened 2025-07-23
Edit user roles

MBausson:
- Delete current roles
- Move the admin role to a boolean attribute `is_admin`


config/locales/en.yml
```diff
@@ -2,6 +2,9 @@ en:
   action_policy:
     flashes:
       unauthorized: You don't have the necessary authorizations
+  boolean:
+    true: "Yes"
+    false: "No"
```
MBausson, line 7: This locale was not working, it needed the double quotes in order to work

MBausson: Yes

config/locales/activerecord.en.yml
```diff
@@ -113,6 +113,8 @@ en:
         bay_id: Bay
         modele_category_id: Category
       theme: Theme
+      user:
+        is_admin: Is administrator
```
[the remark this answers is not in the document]
MBausson: It's already the case : https://github.com/pantographe/my-dcim/blob/6e8a97e78878feeb4c891eda3b2a039a9100ad77/config/locales/activerecord.fr.yml#L274

nicolas-brousse (approved):

PR #450, opened 2025-07-24
Create new user roles

MBausson:
With two roles:
- Reader
- Writer

### DIFF
--- a/db/migrate/20250724082513_add_new_roles_to_users.rb
+++ b/db/migrate/20250724082513_add_new_roles_to_users.rb
@@ -1,0 +1,7 @@
+# frozen_string_literal: true
+
+class AddNewRolesToUsers < ActiveRecord::Migration[8.0]
+  def change
+    add_column :users, :role, :integer
+  end
+end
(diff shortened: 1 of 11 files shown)


db/migrate/20250724082513_add_new_roles_to_users.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class AddNewRolesToUsers < ActiveRecord::Migration[8.0]
+  def change
+    add_column :users, :role, :integer
```
nicolas-brousse, line 5:
```suggestion
    add_column :users, :role, :integer, null: false, default: 0
```

I think we should avoid having null value, and make users `reader`s as default. And made existing users as `writer`s

db/migrate/20250724082513_add_new_roles_to_users.rb
```diff
@@ -0,0 +1,17 @@
...
+
+  enum :role, { reader: 0, writer: 1 }
+end
+
+class AddNewRolesToUsers < ActiveRecord::Migration[8.0]
+  def change
+    add_column :users, :role, :integer, null: false, default: 0
+
+    up_only do
+      MigrationUser.update_all(role: :writer)
```
nicolas-brousse, line 14:
```suggestion
      MigrationUser.reset_column_information
      MigrationUser.update_all(role: :writer)
```

PR #452, opened 2025-07-30
Add policies to models

MBausson:


app/controllers/ports_controller.rb
```diff
@@ -23,10 +25,12 @@ def index
   end
 
   def show
-    @port = Port.find_by_id(params[:id])
+    authorize! @port = Port.find_by_id(params[:id]), with: PortPolicy
```
B-Rass, line 28: What wasn't working than you had to add `, with: PortPolicy`?

B-Rass (changes requested):
I also noticed :

- [x] User can't access "Forgotten password" page
- [ ] Inscription is not accessible

MBausson:
> I also noticed :
> 
>     * [ ]  User can't access "Forgotten password" page
> 
>     * [ ]  Inscription is not accessible

I fixed the reset password page.  
For the registration, I think there shouldn't be this link on the sign in page, it doesn't make sense

app/controllers/frames_controller.rb
```diff
@@ -58,14 +58,16 @@ def update
   end
 
   def sort
+    authorize! to: :sort?
```
nicolas-brousse, line 61: Why specifying the action here is required?

MBausson: This was a mistake, i removed it

app/controllers/data_import_controller.rb
```diff
@@ -4,22 +4,16 @@ class DataImportController < ApplicationController
   skip_before_action :verify_authenticity_token, only: [:ansible]
 
   def index
-    authorize!
+    authorize! with: DataImportPolicy
   end
 
   def ansible
-    authorize!
+    authorize! with: DataImportPolicy
```
MBausson, line 11: I had to use `with:`, because the `default_authorization_policy_class` method wasn't being called by Action Policy 🤷‍♂️

B-Rass (approved):

nicolas-brousse (changes requested):

PR #453, opened 2025-08-06
Imrove search on changelog entries

MBausson:
- Plain text search seems case sensitive.
- There are missing French translations for filters labels
- Filter lists are not sorted

### DIFF
--- a/config/locales/fr.yml
+++ b/config/locales/fr.yml
@@ -905,8 +905,10 @@ fr:
         special_case: Backbone réseau
         comments: Commentaires
         vlans: Vlans
+        actions: Actions
         authors: Auteurs
-        object_types: Types d'objets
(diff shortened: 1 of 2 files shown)


config/locales/fr.yml
```diff
@@ -905,6 +905,8 @@ fr:
         special_case: Backbone réseau
         comments: Commentaires
         vlans: Vlans
+        authors: Auteurs
+        object_types: Types d'objets
```
B-Rass, line 909:
```suggestion
        object_types: Types d'objet
```

B-Rass: Add english translation also

B-Rass (changes requested):

B-Rass (approved):

PR #536, opened 2025-12-17
417317 - Reference moved connections to move projects

nicolas-brousse:
- [ ] What about existing `MovedConnection`s?
- [ ] `Move` shoud be refereed in a step

Lifecycle of MovedConnection around Moves should be looked to be sure not forget some cases.


db/migrate/20251210161815_add_step_reference_to_moved_connections.rb
nicolas-brousse, line 1: Confirm what we do about existing data (don't forget about executed)

nicolas-brousse:
Closed in favor of https://github.com/nanego/my-dcim/pull/662

PR #615, opened 2026-03-19
423456 glpi network equipment sync

infdev-fil:


app/views/power_distribution_units/show.html.erb
nicolas-brousse, line 1: Simply remove glpi related code here

PR #638, opened 2026-04-08
442027 - Keep sort options while choosing columns to display and filters 

B-Rass:


nicolas-brousse (approved):
I did a change. If ok for you then this could be merge.

PR #668, opened 2026-04-27
make changing islet on visualization clean the bay

infdev-fil:


app/views/visualization/rooms/show.html.erb
```diff
@@ -217,6 +217,9 @@ function menuRoomsClick(e) {
 
   // Update page title
   document.querySelector("h1").textContent = visualization_title + " " + islet_name
+
+  // empty bay container
+  $('#bay-container').empty()
```
B-Rass:
```suggestion
  $("#bay-container").empty()
```

app/views/visualization/rooms/show.html.erb
B-Rass, line 1: I also think it could be fine to remove `view`, `bay-id`, `frame-id` from url search parameters

B-Rass:
```suggestion
<script type="text/javascript" nonce="<%= request.content_security_policy_nonce %>">
```

B-Rass: Herb linter recommendation

B-Rass (changes requested):

app/views/visualization/rooms/show.html.erb
B-Rass: Ah and after some thoughts, may be good to not use jQuery here. We want to remove it at the end

app/views/visualization/rooms/show.html.erb
infdev-fil: All of them or just `bay-id` ?

app/views/visualization/rooms/show.html.erb
```diff
@@ -217,6 +217,15 @@ function menuRoomsClick(e) {
...
   // Update page title
   document.querySelector("h1").textContent = visualization_title + " " + islet_name
+
+  // empty bay container
+  const bayContainer = document.getElementById("bay-container")
+  bayContainer.childNodes.forEach((c) => bayContainer.removeChild(c))
+
+  // remove bay container url params
+  const searchParams = new URLSearchParams(window.location.search)
+  ;["bay-id", "view", "frame-id"].forEach(p => searchParams.delete(p))
```
B-Rass, line 227: Why starting your line with `;`?

infdev-fil: Because without it will think we tryed to index `new URLSearchParams(window.location.search)` 
I know for sure tools like [eslint](https://eslint.org/) require those

B-Rass (approved):

PR #692, opened 2026-05-13
Add fields from glpi

infdev-fil:


app/services/glpi_client.rb
```diff
@@ -96,21 +95,20 @@ def get_glpi_id_for(endpoint, serial:)
...
     attributes[:hard_drives] = body["_devices"].present? ? body["_devices"]["Item_DeviceHardDrive"] : {}
     attributes[:memories] = body["_devices"].present? ? body["_devices"]["Item_DeviceMemory"] : {}
+    attributes[:state] = body["states_id"].zero? ? "" : get_state_for(id: body["states_id"])
+
+    group_ids = body["groups_id_tech"].is_a?(Array) ? body["groups_id_tech"] : []
+    attributes[:groups] = group_ids.map { |id| get_group_for(id:) }.to_sentence
+
+    contract_ids = body["_contracts"].is_a?(Array) ? body["_contracts"].pluck("contracts_id") : []
+    attributes[:contracts] = contract_ids.map { |id| get_contract_for(id:) }
+
```
nicolas-brousse: We are agree that these lines are adding extra HTTP requests?
Maybe we should instead have queries that load all groups, states, devicesprocessors, ..., then have the id as key of Hash, then search inside.

Also this list could be load once if needed.
We also could image adding some Rails cache to keep this data a bit of time to avoid those repetitive HTTP requests.

nicolas-brousse: We also could image having this attributes out, and having method inside the object model that do the job. 

```rb
def groups
  get_groups_for_ids(group_ids).map { |g| g["name"] }
end
```

test/services/network_equipment_algori.json
```diff
@@ -50,7 +50,24 @@
     "NetworkPortFiberchannel": []
   },
   "_infocoms": [],
-  "_contracts": [],
+   "_contracts": [
```
nicolas-brousse:
```suggestion
  "_contracts": [
```

config/locales/en.yml
```diff
@@ -433,6 +433,22 @@ en:
...
+      groups: Responsible groups
+      contract: Contract %{type}
+      state: State
+      warranty: Warranty
+      warranty_exprired: Expired
+      disks_count: Disks count
+      disks_capacity: Disks capacity
+      memories_count: Memories count
+      memories_total: Memories total
+      cpus: Cpus
```
nicolas-brousse:
```suggestion
      cpus: CPUs
```

config/locales/en.yml
```diff
@@ -433,6 +433,22 @@ en:
     destroy:
       title: Delete the port type
 
+  glpi:
+    computer:
+      name: Name
+      id: GLPI Id
```
nicolas-brousse:
```suggestion
      id: GLPI ID
```

config/locales/fr.yml
```diff
@@ -479,7 +479,11 @@ fr:
       name: Nom
       id: Id GLPI
```
nicolas-brousse:
```suggestion
      id: ID GLPI
```

app/views/servers/glpi_equipment/show.html.erb
```diff
@@ -18,8 +18,27 @@
...
-          <dt class="pb-2"><%= t("glpi.computer.contact") %></dt>
-          <dd class="mb-0 pb-2 ps-3"><%= @equipment.contact %></dd>
+          <dt class="pb-2"><%= t("glpi.computer.state") %></dt>
+          <dd class="mb-0 pb-2 ps-3"><%= @equipment.state %></dd>
+
+          <dt class="pb-2"><%= t("glpi.computer.warranty") %></dt>
+          <dd class="mb-0 pb-2 ps-3">
+            <% if @equipment.under_warranty? %>
+              <%= l @equipment.last_warranty_end_date %>
+            <% elsif @equipment.warranties.present? %>
```
nicolas-brousse:
```suggestion
            <% elsif @equipment.warranty_exprired? %>
```

I suggest to have this helper instead

app/services/glpi_client.rb
```diff
@@ -97,54 +96,105 @@ def get_glpi_id_for(endpoint, serial:)
...
 
     attributes = body
+    attributes[:_lazy] = {}
+
     attributes[:hard_drives] = body["_devices"].present? ? body["_devices"]["Item_DeviceHardDrive"] : {}
     attributes[:memories] = body["_devices"].present? ? body["_devices"]["Item_DeviceMemory"] : {}
+    attributes[:state] = body["_keys_names"].present? ? body["_keys_names"]["states_id"] : ""
+
+    group_ids = body["groups_id_tech"].is_a?(Array) ? body["groups_id_tech"] : []
+    attributes[:_lazy][:groups] = with_error_handling { group_ids.map { |id| get_group_for(id:) }.to_sentence }
```
nicolas-brousse: I think`.to_sentence` should be called inside view instead here. Don't seems to be the job of the API client.

config/locales/en.yml
```diff
@@ -433,6 +433,22 @@ en:
...
+      serial: Serial number
+      groups: Responsible groups
+      contract: Contract %{type}
+      state: State
+      warranty: Warranty
+      warranty_exprired: Expired
+      disks_count: Disks count
+      disks_capacity: Disks capacity
+      memories_count: Memories count
```
nicolas-brousse:
```suggestion
      memories_count: RAM modules
```

config/locales/en.yml
```diff
@@ -433,6 +433,22 @@ en:
...
+      memories_total: Memories total
```
nicolas-brousse:
```suggestion
      memories_total: RAM total
```

nicolas-brousse (changes requested):

nicolas-brousse (approved):

PR #706, opened 2026-05-20
438460 - Create Move::Connection

B-Rass:


app/models/move.rb
```diff
@@ -10,6 +10,8 @@ class Move < ApplicationRecord
 
   has_one :moves_project, through: :step
 
+  has_many :connections, dependent: :destroy
```
nicolas-brousse, line 13: Add test in Move spec

db/migrate/20260506125949_create_move_connection_model.rb
```diff
@@ -0,0 +1,19 @@
...
+    create_table :move_connections do |t|
+      t.string :cable_name
+      t.string :cable_color
+      t.string :vlans
+
+      t.datetime :executed_at
+
+      t.references :move, null: false, foreign_key: true
+      t.references :port_from, null: false, foreign_key: { to_table: :ports }
+      t.references :port_to, null: false, foreign_key: { to_table: :ports }
```
nicolas-brousse:
```suggestion
      t.references :port_to, null: true, foreign_key: { to_table: :ports }
```

Since it is optional

nicolas-brousse (changes requested):

nicolas-brousse (approved):

PR #717, opened 2026-05-27
451141 - refactor: group CardType by PortType on Server form

nicolas-brousse:
```rb
CardType.find_each.map do |type|
  [type.name, Server.where(id: Card.select(:server_id).where(card_type: type)).pluck(:name)]
end.to_h
```

To get the list of used `CardType` with servers related.

### DIFF
--- a/app/views/servers/_card_fields.html.erb
+++ b/app/views/servers/_card_fields.html.erb
@@ -34,8 +34,7 @@
         <span class="col-12 mt-2 col-sm-4 col-xl-3 ">
           <%= f.label Enclosure.human_attribute_name(:card_type_id), class: "form-label col-form-label-sm p-0" %>
           <%= f.select :card_type_id,
-                       options_for_select(CardType.sorted.map { |card_type| [card_type.name, card_type.id] },
-                                          f.object.card_type_id),
+                       CardTypeDecorator.grouped_by_port_type_options_for_select,
                        {},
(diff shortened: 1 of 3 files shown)


B-Rass (approved):

app/views/servers/_card_fields.html.erb
```diff
@@ -34,8 +34,7 @@
         <span class="col-12 mt-2 col-sm-4 col-xl-3 ">
           <%= f.label Enclosure.human_attribute_name(:card_type_id), class: "form-label col-form-label-sm p-0" %>
           <%= f.select :card_type_id,
-                       options_for_select(CardType.sorted.map { |card_type| [card_type.name, card_type.id] },
-                                          f.object.card_type_id),
+                       CardTypeDecorator.grouped_by_port_type_options_for_select,
```
nanego, line 37: With this new version, the selected value is not the correct saved value, but the first in the list.

nicolas-brousse: We shouldn't have use `grouped_options_for_select` inside the decorator, and simply give the Hash to the `select` helper. It will have add the `selected`.

`select` helper call `grouped_options_for_select` and specify the value without specifying it.
https://github.com/rails/rails/blob/main/actionview/lib/action_view/helpers/tags/select.rb#L19-L32

PR #718, opened 2026-05-28
block moves project step execution when previous unexecuted moves remains

infdev-fil:


app/models/moves_project_step.rb
```diff
@@ -15,6 +15,8 @@ def to_s
   end
 
   def execute!(apply_connections: true)
+    raise unless prev_moves_all_executed?
```
nicolas-brousse: Create a named exception an use it.

app/controllers/moves_project_steps_controller.rb
```diff
@@ -24,11 +24,18 @@ def print
...
-      format.json { head :no_content }
+      respond_to do |format|
+        format.html { redirect_to moves_project_path(@moves_project), notice: t(".flashes.executed") }
+        format.json { head :no_content }
+      end
+    else
+      respond_to do |format|
+        format.html { redirect_to moves_project_path(@moves_project), alert: t(".flashes.prev_moves_not_all_executed") }
+        format.json { head :bad_request }
+      end
```
nicolas-brousse: Remove `if` condition and replace by a rescue of the exception instead

app/models/moves_project_step.rb
```diff
@@ -24,6 +26,10 @@ def executed?
     moves.any?(&:executed?)
   end
 
+  def prev_moves_all_executed?
+    @prev_moves_all_executed ||= Move.where(step: previous_steps, executed_at: nil).none?
+  end
```
nicolas-brousse: Not sure about the memoization here.

app/models/moves_project_step.rb
```diff
@@ -24,6 +26,10 @@ def executed?
     moves.any?(&:executed?)
   end
 
+  def prev_moves_all_executed?
```
nicolas-brousse:
```suggestion
  def prev_moves_executed?
```

nicolas-brousse (changes requested):

app/models/moves_project_step.rb
infdev-fil: I thought that like we check before in controller there were no need to do it twice, but since we throw in `execute!`, the check is only done once

app/models/moves_project_step.rb
```diff
@@ -1,6 +1,9 @@
 # frozen_string_literal: true
 
 class MovesProjectStep < ApplicationRecord
+  class PreviousMovesNotExecutedError < StandardError
+  end
```
nicolas-brousse:
```suggestion
  class PreviousMovesNotExecutedError < StandardError; end
```

app/models/moves_project_step.rb
```diff
@@ -15,6 +18,8 @@ def to_s
   end
 
   def execute!(apply_connections: true)
+    raise PreviousMovesNotExecutedError unless prev_moves_executed?
+
```
nicolas-brousse, line 21: Not tested

app/models/moves_project_step.rb
nicolas-brousse: Could also have been done directly in `execute!` since this is not a big like, but let keep it like this.

nicolas-brousse (changes requested):

spec/models/moves_project_step_spec.rb
```diff
@@ -25,7 +25,24 @@
...
+
+    context "with prev moves executed" do
+      let(:move) { moves(:one) }
+
+      it do
+        expect do
+          move_project_step.execute!
+          move.reload
+        end.to change(move, :executed?).from(false).to(true)
+      end
```
nicolas-brousse:
```suggestion
      it :aggregate_failures do
        move_project_step.moves.each do |move|
          expect do
            move_project_step.execute!
            move.reload
          end.to change(move, :executed?).from(false).to(true)
        end
      end
```

Let's do that to test every moves of the step

nicolas-brousse (changes requested):

nicolas-brousse (approved):

PR #760, opened 2026-06-30
368436 - Enable connection between PDU and Server

B-Rass:

PR #764, opened 2026-07-09
Disable type field update on pdu if there is connectoins

infdev-fil:


moxyom:
Add tests and remove `connections` method when connections (#760) are merged

PR #771, opened 2026-07-15
fix confirm cancel not working on moves execute button

infdev-fil:


nicolas-brousse (changes requested):
Execute button for all step has the same issue

PR #772, opened 2026-07-15
368436 - Some improvments on PDUs

nicolas-brousse:


app/decorators/cable_decorator.rb
```diff
@@ -58,13 +58,7 @@ def draw_port(connection)
 
     if (port = connection&.port)
       port_type = connection&.port_type
-
-      # TODO-PDU: improve this
-      port_type_class = if connection.port.power_distribution_unit_socket?
-                          "portALIM"
-                        else
-                          port_type&.decorated&.css_class_name
-                        end
```
infdev-fil, line 67: If the port_type's name isn't "ALIM" ; even tho `connection.port` is associated with a socket, then the class won't be `"portALIM"`. 

This could happen if the user has associated a non alim port to a `PDU::socket`.
Is it ok ? Should we add a validation on what `PortType` can be associated with a `PDU::Socket` ?

nicolas-brousse: A socket should only have power `PortType`, currently there is only `ALIM` but `ALIM` is not intent to be use for PDU, only for Server. That's why there is changes needed on `PortType` to clean that.

> This could happen if the user has associated a non alim port to a PDU::socket.
Is it ok ? Should we add a validation on what PortType can be associated with a PDU::Socket ?

Should not be the case, and currently couldn't be the case since we show PortType that are only `power` .

PR #775, opened 2026-07-16
464409 - Show glpi empty card on server show

infdev-fil:


app/components/card_empty_data_component.rb
```diff
@@ -11,7 +11,12 @@ def initialize(icon: :slash_circle, text: I18n.t("card_empty_data_component.titl
   def call
     render CardComponent.new(extra_classes: "text-center text-secondary-emphasis") do
       concat(tag.span(class: "bi bi-#{@icon} fs-1 text-secondary text-opacity-25"))
-      concat(tag.h5(@text, class: "card-title mt-3")) if @text.present?
+
+      if content.present?
```
nicolas-brousse:
```suggestion
      if content?
```

nicolas-brousse (changes requested):

PR #778, opened 2026-07-20
368436 - Add power line to PDU and remove name column

infdev-fil:


app/models/power_distribution_unit.rb
```diff
@@ -35,7 +36,14 @@ class PowerDistributionUnit < ApplicationRecord
   before_create :build_circuits_and_sockets_from_type
 
   def should_generate_new_friendly_id?
-    slug.blank? || name_changed?
+    return true if slug.blank? || power_line_changed?
+
+    frame_name = frame&.name
+    frame_name.nil? || !slug.starts_with?(frame.name)
```
nicolas-brousse, line 42:
```suggestion
    return true if slug.blank? || power_line_changed? || frame_id_changed?

    !slug.starts_with?(frame&.name)
```

Does this enough?

db/migrate/20260720131449_add_power_line_to_power_distribution_units.rb
```diff
@@ -0,0 +1,53 @@
+# frozen_string_literal: true
+
+class MigrationFrame < ActiveRecord::Base
+  self.table_name = :frames
+
+  has_many :power_distribution_units, class_name: "MigrationPowerDistributionUnit", foreign_key: :frame_id
+
+  scope :with_pdus, -> { joins(:power_distribution_units).includes(:power_distribution_units).distinct }
```
nicolas-brousse, line 8: Doesn't `includes` redondent with `joins`?

db/migrate/20260720131449_add_power_line_to_power_distribution_units.rb
```diff
@@ -0,0 +1,53 @@
...
+  def up
+    add_column :power_distribution_units, :power_line, :string
+
+    MigrationFrame.reset_column_information
+    MigrationFrame.with_pdus.find_each do |frame|
+      raise "More than 2 pdu for frame with id=#{frame.id}" if frame.power_distribution_units.count > 2
+
+      frame.power_distribution_units.each_with_index do |pdu, i|
+        pdu.power_line = i.even? ? "a" : "b"
+        pdu.save!
```
nicolas-brousse:
```suggestion
        pdu.update!(power_line: i.even? ? "a" : "b")
```

db/migrate/20260720131449_add_power_line_to_power_distribution_units.rb
```diff
@@ -0,0 +1,53 @@
...
+    remove_column :power_distribution_units, :name
+  end
+
+  def down
+    add_column :power_distribution_units, :name, :string
+
+    MigrationPowerDistributionUnit.reset_column_information
+    MigrationPowerDistributionUnit.find_each do |p|
+      p.name = "#{p.frame.name}-#{p.power_line.upcase}"
+      p.save!
```
nicolas-brousse:
```suggestion
      p.update!(name: "#{p.frame.name}-#{p.power_line.upcase}")
```

db/migrate/20260720131449_add_power_line_to_power_distribution_units.rb
nicolas-brousse, line 1: Slugs don't seems to be updated

app/views/power_distribution_units/_form.html.erb
nicolas-brousse, line 1: Update show view as done here

spec/models/power_distribution_unit_spec.rb
```diff
@@ -77,7 +76,7 @@
 
   describe "#build_circuits_and_sockets_from_type" do
     let!(:power_distribution_unit) do
-      described_class.create!(**power_distribution_units(:one).attributes, id: nil, type:, serial_number: "test123456789")
+      described_class.create!(**power_distribution_units(:one).attributes, id: nil, type:, serial_number: "test123456789", power_line: :b)
```
nicolas-brousse, line 79: Really required?

spec/processors/power_distribution_units_processor_spec.rb
```diff
@@ -26,11 +26,11 @@
     let(:islet) { Islet.create!(name: "I1", room:) }
     let(:bay) { Bay.create!(name: "B1", islet:, bay_type: bay_types(:one)) }
     let(:frame) { Frame.create!(name: "F1", bay:) }
-    let(:power_distribution_unit) { PowerDistributionUnit.create!(name: "PDU1", serial_number: "10", **pdu_attr, frame:) }
+    let(:power_distribution_unit) { PowerDistributionUnit.create!(serial_number: "10", **pdu_attr, frame:) }
 
     before do
       power_distribution_unit
-      PowerDistributionUnit.create!(name: "PDU1", serial_number: "11", **pdu_attr, frame: frames(:one))
+      PowerDistributionUnit.create!(serial_number: "11", **pdu_attr, frame: frames(:one), power_line: :b)
```
nicolas-brousse, line 33: Is this important since we test `room_ids` here?

spec/processors/power_distribution_units_processor_spec.rb
```diff
@@ -199,11 +199,11 @@
   describe "when filtering by manufacturer_ids" do
     let(:manufacturer) { Manufacturer.create!(name: "M1") }
     let(:type) { PowerDistributionUnitType.create!(name: "T1", current_type: :three_phase, manufacturer:) }
-    let(:power_distribution_unit) { PowerDistributionUnit.create!(name: "PDU1", serial_number: "10", **pdu_attr, type:) }
+    let(:power_distribution_unit) { PowerDistributionUnit.create!(serial_number: "10", **pdu_attr, type:) }
 
     before do
       power_distribution_unit
-      PowerDistributionUnit.create!(name: "PDU1", serial_number: "11", **pdu_attr, type: power_distribution_unit_types(:one))
+      PowerDistributionUnit.create!(serial_number: "11", **pdu_attr, type: power_distribution_unit_types(:one), frame: frames(:one), power_line: :b)
```
nicolas-brousse, line 206: Is this important since we test `manufacturer_ids` here?

nicolas-brousse (changes requested):

app/models/power_distribution_unit.rb
infdev-fil: When you pass something to `start_with?`, it try to convert to String and i get an error : 
```ruby
"a".starts_with?(nil)
# -> 'String#start_with?': no implicit conversion of nil into String (TypeError)
```

So I don't think it's possible to simplify that way.

db/migrate/20260720131449_add_power_line_to_power_distribution_units.rb
infdev-fil: It's to only retrieve frames that have pdus. Like that, we do not loop over all of them.
Technically, I think `joins` makes an `INNER JOIN`
And `includes` makes an `LEFT JOIN`

spec/models/power_distribution_unit_spec.rb
infdev-fil: We don't update `frame_id` attribute, and like there is a unique validation on `[:frame_id, :power_line]`, it is mandatory to updates all tests to make sure new pdu created pass that validation.

spec/processors/power_distribution_units_processor_spec.rb
infdev-fil: Frame `one` has already a pdu with `power_line: :a`

spec/processors/power_distribution_units_processor_spec.rb
infdev-fil: Frame `one` has already a pdu with `power_line: :a`

db/migrate/20260720131449_add_power_line_to_power_distribution_units.rb
nicolas-brousse: `includes` doesn't make SQL joins. It does an another SQL query.

nicolas-brousse (approved):

PR #779, opened 2026-07-21
368436 Socket unique number per PDU or PDUType

infdev-fil:


app/models/power_distribution_unit/socket.rb
```diff
@@ -11,5 +11,17 @@ class Socket < ApplicationRecord
     has_one :power_distribution_unit, through: :circuit, source: :record, source_type: "PowerDistributionUnit"
 
     validates :number, presence: true
+    validate :number_unique_within_pdu_or_pdu_type
+
+    private
+
+    def number_unique_within_pdu_or_pdu_type
```
nicolas-brousse:
```suggestion
    def number_unique_within_circuit_record
```

spec/models/power_distribution_unit/socket_spec.rb
```diff
@@ -24,4 +24,24 @@
...
+    end
+
+    context "when number is taken" do
+      subject(:socket) do
+        described_class.new(number: 1, circuit_id: 1, port_type_id: 1)
+      end
+
+      it { is_expected.not_to be_valid }
+    end
+  end
```
nicolas-brousse: Move in validation description block

nicolas-brousse (changes requested):

spec/models/power_distribution_unit/socket_spec.rb
```diff
@@ -23,5 +23,25 @@
     it { is_expected.to be_valid }
 
     it { is_expected.to validate_presence_of(:number) }
+
+    describe "number_unique_within_circuit_record" do
+      before { power_distribution_unit_sockets(:one).save }
```
nicolas-brousse: Fixtures are already present in database. We don't need to use save so. What was the issue you've got to do this line?

infdev-fil: Yes you right

nicolas-brousse (approved):

PR #781, opened 2026-07-22
368436 List pdu on bay and frame show page

infdev-fil:
…boFrameComponent


app/views/frames/show.html.erb
nicolas-brousse, line 1: Hide frame column here

nicolas-brousse (changes requested):

nicolas-brousse (approved):

PR #799, opened 2026-08-19
i18n: rework decorator locale keys and calls

nicolas-brousse:
- Move decorators locales in separate file as components, and use same key composition
- Correct missing `human_attribute_name` usage

### DIFF
--- a/app/decorators/bay_decorator.rb
+++ b/app/decorators/bay_decorator.rb
@@ -37,10 +37,10 @@ class BayDecorator < ApplicationDecorator
   def no_frame_warning_icon
     tag.span class: "bay-with-no-frame-warning ms-2" do
       concat(tag.span(class: "bi bi-exclamation-triangle-fill text-warning",
-                      title: I18n.t(".bays.decorator.no_frame_warning_text"),
+                      title: I18n.t("bay_decorator.no_frame_warning_text"),
                       aria: { hidden: true },
                       data: { controller: "tooltip", bs_placement: "right" }))
-      concat(tag.span(I18n.t(".bays.decorator.no_frame_warning_text"), class: "visually-hidden"))
+      concat(tag.span(I18n.t("bay_decorator.no_frame_warning_text"), class: "visually-hidden"))
     end
   end
 end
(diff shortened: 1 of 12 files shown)


app/controllers/concerns/localization.rb
```diff
@@ -11,7 +11,9 @@ module Localization
     def set_locale
       stored_locale = current_user&.locale
       I18n.locale =
-        if stored_locale && I18n.available_locales.include?(stored_locale.to_sym)
+        if params[:locale]
+          params[:locale]
```
B-Rass: I couldn't figure why you needed this change. Can you light me up?

spec/decorators/external_app_record_decorator_spec.rb
```diff
@@ -18,15 +18,15 @@
     context "with external_serial is not present" do
       it { is_expected.to be_a BadgeComponent }
       it { expect(badge.instance_variable_get(:@color)).to eq :danger }
-      it { expect(badge.content).to eq "NON TROUVÉ DANS GLPI" }
+      it { expect(badge.content).to eq "NOT FOUND" }
```
B-Rass: Why is it in english now? Don't we agree to have french here? Or was it by default that we choose that?

app/controllers/concerns/localization.rb
nicolas-brousse: That was for test at first, to easily switch locale by adding `locale=en` in the URL. Maybe we could keep it, otherwise I'll remove it.

spec/decorators/external_app_record_decorator_spec.rb
nicolas-brousse: Since I use `human_attribute_name` instead `I18n.t`. But I didn't check more. It's a bit wierd

app/controllers/concerns/localization.rb
B-Rass: I'm not 100% sure about this, but I'll say to remove it, since we don't need it, and if we do one day, then we'll add it

spec/decorators/external_app_record_decorator_spec.rb
B-Rass: Yes it is

nicolas-brousse: It was an error. Good catch.

PR #808, opened 2026-08-31
456601 prevent card type change if not enough ports

moxyom:


spec/models/card_spec.rb
```diff
@@ -20,6 +20,29 @@
...
+      before do
+        card.card_type = card_types(:two)
+        card.validate
+      end
+
+      it { expect(card).not_to be_valid }
+      it { expect(card.errors.where(:card_type_id, :not_enough_ports).count).to eq(1) }
+    end
+  end
+
```
nicolas-brousse, line 43:
```suggestion
  describe "#ensure_card_type_have_enough_ports" do
    before do
        card.card_type = card_type
        card.validate
      end

    context "when enough ports" do
      let(:card_type) { card_types(:four) }

      it { expect(card).to be_valid }
    end

    context "when not enough ports" do
      let(:card) { cards(:one) }
      let(:card_type) { card_types(:two) }

      it { expect(card).not_to be_valid }
      it { expect(card.errors.where(:card_type_id, :not_enough_ports).count).to eq(1) }
    end
  end

```

app/models/card.rb
```diff
@@ -69,4 +70,13 @@ def set_twin_card
         .update_all({ twin_card_id: nil }) # rubocop:disable Rails/SkipsModelValidations
     end
   end
+
+  private
+
+  def ensure_card_type_have_enough_ports
+    connection_count = ports.joins(:connections).count
```
nicolas-brousse: Prefer adding `has_many :connections` and use `connections.count` here

nicolas-brousse (changes requested):
Error message in French isn't fully translated. I've got the following: `Cards card type ne contient pas assez de ports`

nicolas-brousse (approved):

PR #812, opened 2026-09-02
Manage not found on connection edit

moxyom:


app/controllers/connections_controller.rb
```diff
@@ -62,20 +62,31 @@ def update
...
     from_port = Port.find(params[:connection][:from_port_id])
-    to_port = Port.find(params[:connection][:to_port_id])
 
-    from_port.connect_to_port(to_port,
-                              params[:connection][:cablename],
-                              params[:connection][:color],
-                              params[:connection][:vlans],
-                              params[:connection][:special_case],
-                              params[:connection][:comments])
+    if params[:connection][:to_port_id].presence
```
B-Rass:
```ruby
if params[:connection][:to_port_id].present?
```

or 

```ruby
if to_port_id = params[:connection][:to_port_id].presence
```

app/controllers/connections_controller.rb
```diff
@@ -62,20 +62,31 @@ def update
...
+                                params[:connection][:cablename],
+                                params[:connection][:color],
+                                params[:connection][:vlans],
+                                params[:connection][:special_case],
+                                params[:connection][:comments])
+
+      @to_server = to_port.server
+    else
+      # destroy connections if to_port_id set to null
+      from_port.connections.each do |connection|
```
B-Rass: Why destroying connection? Did you notice case where it was needed?

I don't say it's not a good idea, just I want to know what makes you add this

app/controllers/connections_controller.rb
```diff
@@ -62,20 +62,31 @@ def update
...
+        connection.cable&.destroy
+        connection.destroy
+      end
+
+      from_port.update(vlans: "", color: "", cablename: "")
```
B-Rass: Better update with values from form. With this solution I can't, for example, add a Vlan to a port without connection

B-Rass (changes requested):

PR #813, opened 2026-09-03
Create pdu card crud

moxyom:


app/models/power_distribution_unit/card.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+class PowerDistributionUnit
+  class Card < CardAbstract
+    belongs_to :record, polymorphic: true
+
+    def to_s
+      "Carte #{record} / #{card_type}"
```
nicolas-brousse: Missing translation

app/models/port.rb
```diff
@@ -78,6 +78,8 @@ def frame
       server.frame
     when PowerDistributionUnit::Socket
       circuit.record.frame
+    when PowerDistributionUnit::Card
+      power_distribution_unit_card.record&.frame
```
nicolas-brousse: Do we really need to use a safe navigator here?

app/views/power_distribution_units/_card_fields.html.erb
```diff
@@ -0,0 +1,58 @@
...
+    <div class="row g-2">
+      <span class="col-12 mt-2 col-xl-5 mt-0">
+        <%= f.label :name, class: "form-label" %>
+        <%= f.text_field :name, class: "form-control" %>
+      </span>
+
+      <span class="col-12 mt-2 col-xl-7 mt-0">
+        <%= f.label Card.human_attribute_name(:card_type_id), class: "form-label col-form-label-sm p-0" %>
+        <%= f.select :card_type_id,
+                     CardTypeDecorator.grouped_by_port_type_options_for_select(f.object.card_type_id),
```
nicolas-brousse: We should use a `PowerDistributionUnit::CardDecorator` here and filter cards to show only the ones enabled for PDU on PortType

app/views/power_distribution_units/_card_fields.html.erb
```diff
@@ -0,0 +1,58 @@
...
+                         controller: :select,
+                         select_clear_button_title_value: t("action.clear_all"),
+                       }
+                     } %>
+      </span>
+    </div>
+
+    <div class="row g-2 mt-2">
+      <span class="col-12 mt-2 col-xl-5 mt-0">
+        <%= f.label Card.human_attribute_name(:orientation), class: "form-label col-form-label-sm p-0" %>
```
nicolas-brousse: Wrong model here

app/views/power_distribution_units/_card_fields.html.erb
```diff
@@ -0,0 +1,58 @@
...
+        <%= f.select :orientation,
+                    Card::ORIENTATIONS.map { |o| [Card.human_attribute_name("orientation.#{o}"), o] },
```
nicolas-brousse: Wrong model here

app/views/power_distribution_units/_card_fields.html.erb
nicolas-brousse:
```suggestion
        <%= f.label :orientation, Card.human_attribute_name(:orientation), class: "form-label col-form-label-sm p-0" %>
```

app/views/power_distribution_units/_card_fields.html.erb
```diff
@@ -0,0 +1,58 @@
...
+    <% end %>
+
+    <div class="row g-2">
+      <span class="col-12 mt-2 col-xl-5 mt-0">
+        <%= f.label :name, class: "form-label" %>
+        <%= f.text_field :name, class: "form-control" %>
+      </span>
+
+      <span class="col-12 mt-2 col-xl-7 mt-0">
+        <%= f.label Card.human_attribute_name(:card_type_id), class: "form-label col-form-label-sm p-0" %>
```
nicolas-brousse:
```suggestion
        <%= f.label :card_type_id, Card.human_attribute_name(:card_type_id), class: "form-label col-form-label-sm p-0" %>
```
**Also** wrong model used here

app/models/power_distribution_unit.rb
```diff
@@ -68,7 +73,8 @@ def slug_candidates
     ]
   end
 
-  def build_circuits_and_sockets_from_type
+  def build_extras_from_type
```
nicolas-brousse:
```suggestion
  def build_associations_from_type
```

nicolas-brousse (changes requested):

app/models/port.rb
moxyom: Not that way definitely, with a `.try` I think, cards can belong to a `PDUType` which doesn't have a frame.
Same for `circuit.record`

nicolas-brousse: Why us `try` instead of safe navigator? 🤔 

I think `#frame` is currently only used when `record` is an `PDU` I forgot about `PDUType`. I suggest not using try and safe navigation either for now. If this broke something then we will see how to change it depend the context of the error.

spec/support/shared_examples/models/card.rb
```diff
@@ -0,0 +1,114 @@
...
+    it { is_expected.to have_many(:ports).dependent(:destroy) }
+    it { is_expected.to have_many(:cables).through(:ports) }
+    it { is_expected.to have_many(:connections).through(:ports) }
+  end
+
+  describe "validations" do
+    it { is_expected.to validate_numericality_of(:first_position).only_integer.is_in(0..100).allow_nil }
+  end
+
+  describe "#ensure_card_type_have_enough_ports" do
```
nicolas-brousse: Move in validations

spec/support/shared_examples/models/card.rb
```diff
@@ -0,0 +1,114 @@
...
+
+      execution
+    end
+
+    context "without created ports" do
+      let(:created_port_positions) { [] }
+
+      it { expect(Port).to have_received(:create).exactly(4).times }
+
+      (1..4).each do |position|
```
nicolas-brousse: Use `port_quantity`

spec/support/shared_examples/models/card.rb
```diff
@@ -0,0 +1,114 @@
...
+      allow(card).to receive(:positions_with_ports)
+        .and_return(created_port_positions)
+
+      execution
+    end
+
+    context "without created ports" do
+      let(:created_port_positions) { [] }
+
+      it { expect(Port).to have_received(:create).exactly(4).times }
```
nicolas-brousse: Use `port_quantity`

spec/support/shared_examples/models/card.rb
```diff
@@ -0,0 +1,114 @@
...
+
+    it { expect(positions_with_ports).to eq([0, 1, 2, 3]) }
+  end
+
+  describe "create_missing_ports" do
+    let(:execution) { card.create_missing_ports }
+    let(:port_quantity) { 4 }
+    let(:created_port_positions) { [2] }
+
+    let(:create_base) { { attachable: card, vlans: nil, color: nil, cablename: nil } }
```
nicolas-brousse:
```suggestion
    let(:attributes) { { attachable: card, vlans: nil, color: nil, cablename: nil } }
```

app/views/power_distribution_unit_types/_form.html.erb
```diff
@@ -140,6 +140,39 @@
           <% end %>
         <% end %>
       </fieldset>
+
+      <fieldset data-controller="nested-form" class="mt-4">
```
nicolas-brousse: Any reason to have `mt-4` on PDU types and not in PDUs ?

app/views/power_distribution_unit_types/_form.html.erb
nicolas-brousse, line 1: Maybe we could avoid repetition here and move to a partial this content

app/models/card.rb
```diff
@@ -1,82 +1,36 @@
...
-    end
-  end
+    return if twin_card_id.blank?
 
-  private
-
-  def ensure_card_type_have_enough_ports
-    port_quantity = card_type&.port_quantity || 0
+    twin_card = Card.where(id: twin_card_id).first
+    if twin_card.twin_card_id != id
```
nicolas-brousse: Why having change this line?

app/models/card.rb
```diff
@@ -1,82 +1,36 @@
...
-        twin_card.twin_card_id = id
-        twin_card.save
-      end
-      # Remove potential duplications
-      Card.where(twin_card_id: [id, twin_card_id])
-        .where.not(id: [id, twin_card_id])
-        .update_all({ twin_card_id: nil }) # rubocop:disable Rails/SkipsModelValidations
-    end
-  end
+    return if twin_card_id.blank?
```
nicolas-brousse:
```suggestion
    return unless twin_card_id
```

ids can't be an empty string, but only nil or object

nicolas-brousse (changes requested):

B-Rass:
It misses some translations : 

image

B-Rass:
image


Don't pre-select card type

app/views/power_distribution_unit_types/_form.html.erb
moxyom: Because of the container, pdu_type's nested fields are in regular `div.col`, while pdu's nested fields are in a `div.d-flex.flex-column.row-gap-4`

app/models/card.rb
moxyom: It better represent what is done, and remove unecessary check (`twin_card.present?`).
Maybe use `find_by` instead of `where.first` to express that twin card must exist ?

app/views/power_distribution_unit_types/_form.html.erb
moxyom: move the nested-form fieldset ?

app/decorators/power_distribution_unit_card_decorator.rb
```diff
@@ -0,0 +1,18 @@
...
+    def card_type_grouped_by_port_type_options_for_select(selected = nil)
+      grouped_card_types = PortType
+        .usable_by_pdu
+        .includes(:card_types)
+        .map do |port_type|
+          [port_type.to_s, port_type.card_types.map { |card_type| [card_type.to_s, card_type.id] }]
+      end
+
+      grouped_options_for_select(grouped_card_types, selected)
+    end
```
nicolas-brousse:
```suggestion
    def card_type_grouped_by_port_type_options_for_select
      PortType
        .usable_by_pdu
        .includes(:card_types)
        .map do |port_type|
          [port_type.to_s, port_type.card_types.map { |card_type| [card_type.to_s, card_type.id] }]
      end
    end
```

This may work

app/models/card.rb
nicolas-brousse: You seem not to look the same informations than the original condition here. Let keep the original to avoid surprise, since the goal of this PR is not to optimize existing Card but abstract the part that we use for PDU Cards

spec/decorators/power_distribution_unit/card_decorator_spec.rb
nicolas-brousse, line 1: Take inspiration of `BayDecorator` spec here, since we'll do not pass the selected value

nicolas-brousse (changes requested):

app/views/power_distribution_unit_types/_form.html.erb
nicolas-brousse: Thanks for the info

B-Rass (commented):
- [ ] Port on PDU have class `no_client`, they shouldn't
- [ ] PDU card connections should appear in "Connections" tab
- [ ] On visualization/frame#show, when a connection is added, reload the PDU visualization (look how it's done for server)
- [ ] Debug "Draw connection" when a pdu card port is connected

app/decorators/power_distribution_unit_card_decorator.rb
nicolas-brousse, line 1: Should be named `PowerDistributionUnit::CardDecorator` since model is named `PowerDistributionUnit::Card`

app/views/power_distribution_unit_types/_form.html.erb
nicolas-brousse: Yes, they are the same block right?

app/views/power_distribution_units/_visualization.html.erb
```diff
@@ -27,5 +27,22 @@
...
+            <%= ports_by_card(port_type: card.card_type.try(:port_type),
+                              port_quantity: card.card_type.try(:port_quantity),
+                              card_id: card.id,
+                              ports_data: card.ports,
+                              is_pdu_card: true,
+                              twin_card_used_ports: nil) %>
+          <% end %>
+        </span>
+      <% end %>
+    </ul>
```
nicolas-brousse, line 46: Preview is missing in PDU Type

app/decorators/power_distribution_unit_card_decorator.rb
moxyom: I also renamed SocketDecorator

